Issue #9, opened 2022-09-30
cuRAND fails when generating an odd number of values


coreylowman:
Apparently this is all expected according to their documentation. This crate should find away around this short coming of their host api.

M1ngXU:
maybe generate ((n + 1) / 2) * 2 number then?
if the number is even, then n = 2m => ((2m + 1) / 2) * 2 = m * 2 = 2m
if the number is odd,  then n = 2m + 1 => ((2m + 1 + 1) / 2) * 2 = (m + 1) * 2 = 2m + 2 (even)

or just use if then else?

coreylowman:
Okay I tried some different logic for if the number of values is odd, but got some weird alignment errors.

A really hacky way I might move forward with is to ALWAYS allocate an even number of values. So in the case where you have an array with an odd number of values, there would be an extra value hanging out at the end unused to round it out.

M1ngXU:
so basically generate if n is odd, then generate n + 1 numbers and truncate the vec length (didn’t find the location in the code, so not sure what the output is). but i’d say that that’s a good solution

coreylowman:
Yeah exactly - will probably want to add `fn malloc_async_even` to result.rs (https://github.com/coreylowman/cudarc/blob/main/src/driver/result.rs#L266), and then move to using that here https://github.com/coreylowman/cudarc/blob/main/src/cudarc.rs#L228

M1ngXU:
ok - do you think it's worth for me to look into the code, because i understand basically nothing 😅

coreylowman:
If you have time and interest, sure! I can help review code or answer questions.

M1ngXU:
ok thx! i'll have a look at it in the next few days. any recommendations of what to read before/where to start with?

coreylowman:
Relevant CUDA documentation:
- cuRAND host api overview: https://docs.nvidia.com/cuda/curand/host-api-overview.html#host-api-overview
- cuRAND host api docs: https://docs.nvidia.com/cuda/curand/group__HOST.html#group__HOST
- CUDA Driver api allocation function: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MALLOC__ASYNC.html#group__CUDA__MALLOC__ASYNC_1g13413273e84a641bce1929eae9e6501f

The functions that have a requirement on even number of values are the log normal & normal generation functions:
[...]

PR #13, opened 2022-11-02
gen_odd

M1ngXU:
probably a bad solution for #9

### DIFF
--- a/examples/devices.rs
+++ b/examples/devices.rs
@@ -1,5 +1,3 @@
-#![feature(generic_associated_types)]
-
 use cudarc::prelude::*;
 use std::rc::Rc;
 
--- a/examples/kernels.rs
+++ b/examples/kernels.rs
@@ -1,4 +1,5 @@
-use cudarc::{jit::*, prelude::*};
+#![feature(generic_const_exprs)]
+use cudarc::{jit::*, prelude::*, arrays::{FixedSizeArray, Array}};
 use std::{marker::PhantomData, rc::Rc};
 
 struct Cpu;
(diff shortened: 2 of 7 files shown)


examples/devices.rs
```diff
@@ -1,5 +1,3 @@
-#![feature(generic_associated_types)]
-
```
M1ngXU, line 2: https://blog.rust-lang.org/2022/10/28/gats-stabilization.html#:~:text=As%20of%20Rust%201.65%2C%20which%20is%20set%20to%20release%20on%20November%203rd

examples/kernels.rs
```diff
@@ -1,4 +1,5 @@
-use cudarc::{jit::*, prelude::*};
+#![feature(generic_const_exprs)]
```
M1ngXU, line 1: this is required to compile, even though no const generics are used directly ...

src/arrays.rs
```diff
@@ -1,9 +1,77 @@
...
+
 pub trait NumElements {
     const NUMEL: usize;
     type Dtype;
 }
 
-pub type Array<N, const A: usize> = [N; (A + 1) / 2 * 2];
+// TODO change names?
+/// An [Array] with `A` as the "real" length but always an even "actual" length
+pub type FixedSizeArray<N, const A: usize> = Array<N, A, {(A + 1) / 2 * 2}>;
```
M1ngXU, line 15: `FixedSizeArray` or `FixedSize**D**Array`?

M1ngXU: ok, can't edit it (lol), so is it `FixedSizeArray` or `FixedSizedArray`?

examples/kernels.rs
coreylowman: Ahh I see just because the crate itself uses them in pub api?

M1ngXU: because the return type of one used function uses const generics. if you remove this line you'll get weird type mismatch errors

M1ngXU:
this doesn't make sense with cudnn anymore. this will probably just be an annoying issue - or we should do the same with tensors

PR #14, opened 2022-11-02
replaced all rc by arc

### LINKED ISSUE #10: Use Arc instead of Rc for sharing between threads

M1ngXU:
closes #10


M1ngXU:
must be *VERY* careful if 2 simultanious writes to the same gpu memory location is possible - didn't check the docs, but this wouldn't work on cublas.

PR #16, opened 2022-11-08
Cudnn support

M1ngXU:
currently only `Tensor4D`, `Activation` and `Convolution`, both forward and backward


M1ngXU:
should the tensordata struct be changed to something like cudaslice?

src/cudnn/tensor/descriptor.rs
```diff
@@ -0,0 +1,66 @@
...
+pub(crate) const TENSOR_FORMAT: cudnnTensorFormat_t = cudnnTensorFormat_t::CUDNN_TENSOR_NCHW;
+/// A descriptor of a tensor. This can be reused as all tensors are fully packed
+/// and in NCHW layout.
+///
+/// This descriptor is destroyed when it is dropped.
+///
+/// # See also
+/// <https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnTensorDescriptor_t>
+/// <https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnDestroyTensorDescriptor>
+pub struct TensorDescriptor<T, const N: usize, const C: usize, const H: usize, const W: usize> {
```
coreylowman, line 15: Yeah this can be changed to hold runtime values instead of const generics

src/cudnn/custom_kernels/f_kernels.cu
```diff
@@ -0,0 +1,55 @@
+// every function must end with `f32` and only accept `float`s; then the same function with `f64` and `double`s will be generated
+
+extern "C" __global__ void recip_with_scale_f32(float *out, const float *a, const float *a_scale, size_t numel)
```
coreylowman, line 3: whoa these are awesome! especially that compile_custom_kernels.sh file. :rocket:

coreylowman:
This does seem to fit more at the device level in dfdx, especially since we'll need a layer there regardless to conform to the kernel traits over there.

M1ngXU:
@coreylowman can you take over these changes? i am only availabe tomorrow for very few hours and maybe a bit on friday, but i could also do them in 3.5 weeks

M1ngXU:
hmm, lots of merge conflicts

src/cudnn/tensor/descriptor.rs
M1ngXU: ok, but should this runtime panic (asserts) or just be unsafe?

M1ngXU: (this is most likely affecting every module in cudnn 😅)

src/cudnn/custom_kernels/f_kernels.cu
M1ngXU: thanks :)
(though this doesn’t work on windows for me lol)

src/cudnn/activation/mode.rs
```diff
@@ -0,0 +1,37 @@
...
+                f64::MAX
+            }
+        }
+    };
+}
+
+impl_activation_mode!(Sigmoid: CUDNN_ACTIVATION_SIGMOID);
+impl_activation_mode!(Relu: CUDNN_ACTIVATION_RELU);
+impl_activation_mode!(Tanh: CUDNN_ACTIVATION_TANH);
+impl_activation_mode!(Elu: CUDNN_ACTIVATION_ELU);
```
coreylowman, line 37: i'm thinking we don't need to add this activation forward for dfdx since there are so little in cudnn. we can just write custom kernels for them (they are really easy to write)

src/cudnn/conv/filter/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::cudnn::sys::*;
+use crate::prelude::*;
+
+/// A convolution filter. Cloning this filter only clones the
+/// pointer to the [FilterDescriptor] and [Tensor4DData] and thus increases the
+/// reference count.
+pub struct Filter<T, const C_OUT: usize, const C_IN: usize, const H: usize, const W: usize> {
+    descriptor: Rc<FilterDescriptor<T, C_OUT, C_IN, H, W>>,
+    data: Tensor4DData<T, C_OUT, C_IN, H, W>,
+}
```
coreylowman, line 19: It's weird how they have a separate descriptor for filters, I wonder why they don't just use a Tensor descriptor?

src/cudnn/tensor/descriptor.rs
coreylowman: runtime panics on creation IMO

src/cudnn/tensor/operation/reduce.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+impl_reduce_op!(ReduceOperationAdd: CUDNN_REDUCE_TENSOR_ADD);
+impl_reduce_op!(ReduceOperationMul: CUDNN_REDUCE_TENSOR_MUL);
+impl_reduce_op!(ReduceOperationMin: CUDNN_REDUCE_TENSOR_MIN);
+impl_reduce_op!(ReduceOperationMax: CUDNN_REDUCE_TENSOR_MAX);
+impl_reduce_op!(ReduceOperationAMax: CUDNN_REDUCE_TENSOR_AMAX);
+impl_reduce_op!(ReduceOperationAvg: CUDNN_REDUCE_TENSOR_AVG);
+impl_reduce_op!(ReduceOperationNorm1: CUDNN_REDUCE_TENSOR_NORM1);
+impl_reduce_op!(ReduceOperationNorm2: CUDNN_REDUCE_TENSOR_NORM2);
+impl_reduce_op!(ReduceOperationMulNoZeros: CUDNN_REDUCE_TENSOR_MUL_NO_ZEROS);
```
coreylowman, line 71: Yeah this will all be super useful. Let's move to using runtime shapes though as mentioned in other places, I can help with this

src/cudnn/tensor/descriptor.rs
coreylowman, line 15: We also need to support up to 6d tensors, which it looks like you can do pretty easily with the tensor descriptors, this can be tracked at runtime like everything else 😄

M1ngXU: so TensorNd = TensorDescriptorNd([Axes]) + CudaSlice?

Issue #20, opened 2022-12-15
More parameters for launching kernels


M1ngXU:
sometimes the current limit (4 or 5) of parameters is not enough; can we have tuples implement the `IntoKernelParam` trait? (maybe `IntoKernelParam` also specifies the size of a parameter, so a tuple of size 2 would have the double size)

PR #24, opened 2022-12-16
Adding Send/Sync impls and impl launch with bigger tuples

### LINKED ISSUE #19: impl Send/Sync for CudaDevice
```rust
unsafe impl Send for CudaDevice {}
unsafe impl Sync for CudaDevice {}
```
is all that is needed.

It doesn't work out of the box because CudaDevice has raw pointers from the sys api.

coreylowman:
Resolves #19 and #20

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -164,6 +164,9 @@ pub struct CudaSlice<T> {
     pub(crate) host_buf: Option<Vec<T>>,
 }
 
+unsafe impl<T: Send> Send for CudaSlice<T> {}
+unsafe impl<T: Sync> Sync for CudaSlice<T> {}
+
 impl<T> CudaSlice<T> {
     /// Number of elements in the slice
     pub fn len(&self) -> usize {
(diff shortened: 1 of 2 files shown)


src/device.rs
```diff
@@ -533,6 +545,9 @@ impl_launch!([A, B], [0, 1]);
 impl_launch!([A, B, C], [0, 1, 2]);
 impl_launch!([A, B, C, D], [0, 1, 2, 3]);
 impl_launch!([A, B, C, D, E], [0, 1, 2, 3, 4]);
+impl_launch!([A, B, C, D, E, F], [0, 1, 2, 3, 4, 5]);
+impl_launch!([A, B, C, D, E, F, G], [0, 1, 2, 3, 4, 5, 6]);
+impl_launch!([A, B, C, D, E, F, G, H], [0, 1, 2, 3, 4, 5, 6, 7]);
```
M1ngXU, line 550: yay! maybe we could make tuples implement intokernelparam by allowing each object implementing IntoKernelParam to have a dynamic size. but 8 parameters are a great start for now

Issue #18, opened 2022-12-14
cudnn support


coreylowman:
See initial discussion in #1  and #16 for wip

On thread safety:
> The cuDNN library is thread-safe. Its functions can be called from multiple host threads, so long as the threads do not share the same cuDNN handle simultaneously.
[Source](https://docs.nvidia.com/deeplearning/cudnn/developer-guide/index.html#thread-safety)

coreylowman:
Re thread safety I think if we enforce that the handles are `Arc<Mutex<CudnnHandle>>`, we should be fine

M1ngXU:
or just let each thread spawn a new handle (no data is associated to a handle)

M1ngXU:
> @M1ngXU let's split the cudnn pr into a couple separate PRs, so we just have what we need for dfdx:
> 
> 1. Add sys.rs/bindgen.sh/result.rs & link to cudnn in build.rs (you already have this in the big PR, just need to split it out)
>    
>    1. i think to start we just need to create/destroy handle, cudnnSetStream
> 2. Add tensor descriptors cudnnCreateTensorDescriptor, cudnnDestroyTensorDescriptor, cudnnSetTensorNdDescriptor
> 3. Add reduction support
> 4. Add conv2d support
> 5. Add pool2d support

ok, i'll do number one now

PR #43, opened 2023-01-09
Cudnn base

M1ngXU:
https://github.com/coreylowman/cudarc/issues/18#issuecomment-1375928870


src/lib.rs
```diff
@@ -22,4 +23,5 @@ pub mod prelude {
         AsKernelParam, CudaDevice, CudaDeviceBuilder, CudaSlice, DriverError, LaunchAsync,
         LaunchConfig, ValidAsZeroBits,
     };
+    pub use crate::cudnn::*;
```
coreylowman, line 26: Let's not include this in the prelude for now

src/cudnn/bindgen.sh
```diff
@@ -0,0 +1,17 @@
...
+  --whitelist-function="^cudnn.*" \
+  --default-enum-style=rust \
+  --no-doc-comments \
+  --with-derive-default \
+  --with-derive-eq \
+  --with-derive-hash \
+  --with-derive-ord \
+  --size_t-is-usize \
+  --use-core \
+  wrapper.h -- --include-directory "/mnt/e/Max/cudnn/include" --include-directory "/mnt/c/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include" \
```
coreylowman, line 16:
```suggestion
  wrapper.h -- -I/usr/local/cuda/include \
```
to keep standard with other bindgens. I guess this doesn't work on windows though?

build.rs
```diff
@@ -5,4 +5,6 @@ fn main() {
     println!("cargo:rustc-link-lib=dylib=cudart");
     println!("cargo:rustc-link-lib=dylib=cublas");
     println!("cargo:rustc-link-lib=dylib=cublasLt");
+    println!("cargo:rustc-link-search=native={}", std::env::var("CUDNN_LIB").expect("Failed to read environmental variable `CUDNN_LIB`. Set `CUDNN_LIB` to `path/to/cudnn/lib`."));
```
coreylowman, line 8: Do you know why this is needed for cudnn but not other libs?

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target
 /Cargo.lock
+cudnn.log
```
coreylowman, line 3: Is this something cudnn creates?

coreylowman (commented):
Looks great! Will approve after cleaning up some of the paths that are specific to your machine it looks like

src/lib.rs
M1ngXU: ok

src/cudnn/bindgen.sh
M1ngXU: hmm, this is in wsl since bindgen doesn't work on windows for me

build.rs
M1ngXU: because cudnn is installed separately and is not in the folder where cuda resides

.gitignore
M1ngXU: i have an env var to log all cudnn calls, i can remove that if necessary

build.rs
coreylowman: Ooooh did not realize that

src/cudnn/result.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for CudnnError {}
+
+// No sync because of https://docs.nvidia.com/deeplearning/cudnn/developer-guide/index.html#thread-safety
+pub struct CudnnHandle(pub(crate) cudnnHandle_t);
```
coreylowman, line 35: This needs to keep a reference to Arc<CudaDevice> to make sure this doesn't outlive device:
```suggestion
pub struct CudnnHandle {
    pub(crate) handle: cudnnHandle_t,
    pub(crate) device: Arc<CudaDevice>,
}
```

And then `fn create` should take an `Arc<CudaDevice>`.

Did a similar thing with blas if you want an example

build.rs
coreylowman: So we'll probably need to wrap this functionality with a feature if it's not installed by default with cuda. Will investigate a bit on my dev machine to see if it's installed/what the process is like. For dfdx I wonder if we should require cudnn, or fallback to custom kernels if people don't have cudnn installed

M1ngXU: does something like pytorch or tensorflow work with cuda but without cudnn? i don't think so and i don't see any reason why someone could install cuda but not cudnn ...

src/cudnn/result.rs
M1ngXU: ok, didn't think of that 😅 

by the way, the test "launch with 16 bit" fails for me with an buffer overflow; are you having the same issue?

coreylowman: I haven't, I can check again

build.rs
coreylowman: I think pytorch bundles cuda/cudnn libraries together? Not completely sure tbh. If cudnn is installed by default then i don't think we need feature flag. We can just require it but I want to go through install process to see how annoying it is first

src/cudnn/result.rs
M1ngXU: ok thanks

build.rs
M1ngXU: not sure about pytorch, but it is seperate in tensorflow (you have to install everything yourself anyway)...

coreylowman:
Closing since this was merged in #97

Issue #57, opened 2023-01-19
ValidAsZeroBits for tuples


M1ngXU:
tuples where all elements are `ValidAsZeroBits` are `ValidAsZeroBits` too

coreylowman:
Is the use case for allocating slices where the elements are tuples?

M1ngXU:
yes

Issue #56, opened 2023-01-19
copy from `CudaView`


M1ngXU:
Sometimes you only need a specific part of a `CudaSlice` and copying from a `CudaView` would be very useful (and why can you only slice with a `RangeFrom`?)

coreylowman:
Can you give example of copying in that manner?

I initially did RangeFrom to make it clear that there is no concept of length for CudaViews. Perhaps with what you are thinking about (copying from CudaView) we could add length and then make the range more generic.

M1ngXU:
if i have a large CudaSlice but only want to copy let's say 100 elements out of it (first 100), then you would have to copy the whole slice and slice the slice you just copied out ...

coreylowman:
This is needed for the stack operation in dfdx - notably you allocate enough space for all the tensors you are stacking, then copy values into sub views

Issue #55, opened 2023-01-19
`sync_copy_from` - also accept vector


M1ngXU:
A function that also accepts a vector would be great as the cudaslice-sizes might be dynamic and it then requires more boilerplate code to actually create a correctly sized slice.

M1ngXU:
or actually returning a vector would be great too

coreylowman:
Does there is `dev.take_async` which takes a Vec, and `sync_release` for returning a vec, do those work?

M1ngXU:
hmm, but what if you only want to copy it out and not destroy the slice? also, i don't think that the function names are very clear ...

coreylowman:
Yeah good point. Any suggestions for naming? Definitely open to suggestions!

M1ngXU:
`sync_copy_into_vec` 😅

M1ngXU:
maybe add the documentation label & rename this issue (or close this one and create a new one?)

PR #59, opened 2023-01-21
public alloc, ValidAsZeroBits for tuples, CudaSlice slicing with any range

### LINKED ISSUE #54: alloc (without memsetting to 0) should be public
can this function be public?
https://github.com/coreylowman/cudarc/blob/2785abba641f3bff4234e0aa70542deab6f67c86/src/driver/safe.rs#L388

Sometimes i know that i'll overwrite/initialize the memory, but memsetting 2GB-3GB before is rather a waste of time.

M1ngXU:
`CudaView` should have the same operations as `CudaSlice` (like slicing, copying, ...), though this would be a seperate PR. This closes #54, #55 (though a better name is required), #56 (partially) and #57.

### DIFF
--- a/src/driver/safe.rs
+++ b/src/driver/safe.rs
@@ -156,6 +156,7 @@ use super::{result, sys};
 use crate::nvrtc::Ptx;
 
 use alloc::ffi::{CString, NulError};
+use core::ops::{Bound, RangeBounds};
 use spin::RwLock;
 use std::{collections::BTreeMap, marker::Unpin, pin::Pin, sync::Arc, vec::Vec};
 
@@ -215,7 +216,7 @@ impl<T> CudaSlice<T> {
 
     /// Allocates copy of self and schedules a device to device copy of memory.
     pub fn clone_async(&self) -> Result<Self, DriverError> {
-        let dst = unsafe { self.device.alloc(self.len) }?;
+        let dst = unsafe { self.device.alloc_async(self.len) }?;
         unsafe {
             result::memcpy_dtod_async(
                 dst.cu_device_ptr,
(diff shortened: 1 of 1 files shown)


src/driver/safe.rs
```diff
@@ -273,15 +302,12 @@ impl<T> CudaSlice<T> {
...
-            })
-        } else {
-            None
-        }
+    pub fn try_slice(&self, range: impl RangeBounds<usize>) -> Option<CudaView<'_, T>> {
+        range.bounds(..self.len()).map(|(start, end)| CudaView {
+            ptr: self.cu_device_ptr + (start * std::mem::size_of::<T>()) as u64,
+            slice: self,
+            len: 1 + end - start,
+        })
```
coreylowman, line 310: This is super cool, nice helper trait!!

src/driver/safe.rs
```diff
@@ -1230,7 +1314,9 @@ extern \"C\" __global__ void sin_kernel(float *out, const float *inp, size_t num
 
         for i in 0..5 {
             let a_sub = a_dev.try_slice(i * 2..).unwrap();
+            assert_eq!(a_sub.len, 10 - 2 * i);
             let mut b_sub = b_dev.try_slice_mut(i * 2..).unwrap();
+            assert_eq!(b_sub.len, 10 - 2 * i);
```
coreylowman, line 1319: Want to adjust this test to use `i*2..i*2+2`?

coreylowman (approved):

src/driver/safe.rs
M1ngXU: hmm, this is actually an existing test where i just added to test the length (while this should probably be seperated ...). But what exactly do you mean? the slicing with i*2..i*2+2; if yes, why?

coreylowman: Mainly want something to test that the length of CudaView is properly used. I think this can be done in #56

PR #60, opened 2023-01-21
Implement ValidAsZeroBits for bool

ViliamVadocz:
This change is needed to implement https://github.com/coreylowman/dfdx/issues/306 because the output of comparison operations is a boolean, and for CUDA we want to initialize the output tensor beforehand.

I believe it is sound, as according to the links below, `0` is guaranteed to be interpreted as `false` in both Rust and C++.

- https://doc.rust-lang.org/reference/types/boolean.html
- https://stackoverflow.com/questions/19351483/how-is-a-bool-represented-in-memory


M1ngXU:
makes sense, i also verified that 0 is false on cuda

coreylowman (approved):
Nice, thanks!

PR #64, opened 2023-01-23
"generic" `CudaSlice`-slicing

### LINKED ISSUE #56: copy from `CudaView`

M1ngXU:
Closes #56

### DIFF
--- a/src/driver/safe.rs
+++ b/src/driver/safe.rs
@@ -189,17 +189,17 @@ pub use result::DriverError;
 /// Unfortunately, `&CudaSlice<T>` can **still be mutated
 /// by the [CudaFunction]**.
 #[derive(Debug)]
-pub struct CudaSlice<T> {
+pub struct CudaSlice<T, D: GetDevice> {
     pub(crate) cu_device_ptr: sys::CUdeviceptr,
     pub(crate) len: usize,
-    pub(crate) device: Arc<CudaDevice>,
+    pub(crate) device: D,
     pub(crate) host_buf: Option<Pin<Vec<T>>>,
 }
 
-unsafe impl<T: Send> Send for CudaSlice<T> {}
-unsafe impl<T: Sync> Sync for CudaSlice<T> {}
+unsafe impl<T: Send, D: GetDevice> Send for CudaSlice<T, D> {}
+unsafe impl<T: Sync, D: GetDevice> Sync for CudaSlice<T, D> {}
 
-impl<T> CudaSlice<T> {
+impl<T, D: GetDevice> CudaSlice<T, D> {
     /// Number of elements in the slice
     pub fn len(&self) -> usize {
         self.len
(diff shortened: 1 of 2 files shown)


M1ngXU:
One question is how to share the host buffer; but this could be another issue?

src/driver/safe.rs
```diff
@@ -189,17 +189,17 @@ pub use result::DriverError;
 /// Unfortunately, `&CudaSlice<T>` can **still be mutated
 /// by the [CudaFunction]**.
 #[derive(Debug)]
-pub struct CudaSlice<T> {
+pub struct CudaSlice<T, D: GetDevice> {
```
coreylowman, line 192: What's the reasoning behind adding this generic to CudaSlice? This is a pretty big break in terms of user interface, and would require a lot of documentation to explain what users should put this generic as.

I think for copying to/from either CudaSlice/CudaView, you should just need to swap the relevant functions to accept a generic of type `DevicePtr` or `DevicePtrMut`. You can add a `len()` method to those traits as well.

M1ngXU: well, D is either `Arc<CudaDevice>` or any borrowing of this (like `&&mutArc<CudaDevice>` would be a reference slice of a mutable slice). you can have a look at the trait definition. also, you can only “create” a cudaslice that has `CudaSlice<T, Arc<CudaDevice>>` (when allocating).

M1ngXU: similar to `OwnedTape` and `NoneTape`; `OwnedType` being `Arc<CudaDevice>` and `NoneTape` just a borrowed `Arc<CudaDevice>`

coreylowman: I'm just not seeing the benefit the additional generic is bringing along with the added complexity. I'd like to keep the CudaSlice struct as simple as possible for people to use, and this complicates it quite a bit. See if you can do this ticket without adding the generic, and if you run into some blocker let's revisit

M1ngXU: hmm ok. So what if there only was `CudaSlice<T>` and `CudaView<T, D>`, where `D` will always be a borrowed form of `Arc<CudaDevice>`? And also a trait to implement all `CudaSlice` functions for `CudaView`?

coreylowman: CudaView right now doesn't store the CudaDevice separately - the main thing is that they borrow a CudaSlice. Definitely think we can add a trait to impl all the same functions for CudaSlice and CudaView! That is what I was suggesting with using DevicePtr and DevicePtrMut for the CudaDevice methods that currently only accept CudaSlice

M1ngXU: well, it borrows CudaSlice, but the only thing that is necessary is CudaDevice; that's why i directly borrow Arc<CudaDevice> and was able to put this into a single struct. how would you handle the slicing of a CudaView?

coreylowman: Agree with you data wise that all that is needed is CudaDevice, but it is necessary safety wise to borrow CudaSlice so they don't get dropped while the views still exist.

Easiest thing I can think of would be to add the try_slice methods to them:

```rust
impl<'a, T> CudaView<'a, T> {
    pub fn try_slice(&self, ...) -> Self {
        ...
    }
}

impl<'a, T> CudaViewMut<'a, T> {
    pub fn try_slice_mut(&self, ...) -> Self {
        ...
    }
}
```

M1ngXU: but the CudaDevice is borrowed from the CudaSlice, so the CudaSlice be dropped (or else the pointer is dangling)
also, with your methods, wouldn't it be possible to have multiple mutable references to the same region of memory (idk if that's an issue though)

coreylowman: Yeah I don't think that's an issue re multiple mutable views at a time - that's not needed for dfdx at least!

M1ngXU: but if you accidently pass 2 mutable views to the same function that actually needs 2 mutable views (=> race condition)?

coreylowman: Sorry, I mean it is an issue (race condition), but it's not a requirement of the API to make that possible (i.e. we can only allow 1 mutable view at a time).

M1ngXU: but how would you prevent this?

coreylowman: Rust's borrow checker will prevent this with lifetimes 😀

M1ngXU: ah ok, forgot that a mutable view on a reference view doesn’t make sense 😅 if i have some time i’ll do that

coreylowman:
Closing this as I resolved #56 in a recent PR

PR #76, opened 2023-02-21
Adding cuda Profiler API.

Narsil:
In order to debug/profile `dfdx` apps, it's useful to be able to simply
define profiler start/stop.

And then `nsys profile -c cudaProfilerApi /path/to/binary`.

Could be behind a feature flag since this is a debug only feature.

Used `allowlist` instead of `whitelist` because of newer bindgen.


src/profiler/safe.rs
```diff
@@ -0,0 +1,11 @@
...
+
+pub use result::DriverError;
+
+pub fn start() -> Result<(), DriverError> {
+    unsafe { sys::cuProfilerStart() }.result()
+}
+
+pub fn stop() -> Result<(), DriverError> {
+    unsafe { sys::cuProfilerStop() }.result()
+}
```
coreylowman, line 11: It seems like these could be moved into the driver module instead of creating a new profiler module?

And then after that lets rename to start_profiler() and stop_profiler()

src/profiler/wrapper.h
```diff
@@ -0,0 +1 @@
+#include "cudaProfiler.h"
```
coreylowman, line 1: Could always just add this header to the driver wrapper.h if its needed

src/profiler/safe.rs
Narsil: Well this modifies `wrapper.h`. Since you would need to include `cudaProfiler.h`.

I was hesitant to do it that way, but I could definitely do it.

src/profiler/wrapper.h
Narsil: Ok will do that.

Narsil: Is it ok if I use `allowlist` instead of `whitelist` ? (bindgen version)

Narsil: Also I don't have cuda 11.3 locally but 11.8, any counterindication to use that ? (It does create a lot of unecessary changes)

coreylowman: Yes I think should be fine for both (I was using old bindgen which they made big updates too).

src/driver/safe.rs
```diff
@@ -1121,6 +1121,27 @@ impl CudaDeviceBuilder {
     }
 }
 
+/// Enables profile collection by the active profiling tool for the current context. If profiling is already enabled, then profiler_start() has no effect.
+/// ```
+/// use cudarc::profiler;
```
coreylowman, line 1126:
```suggestion
/// ```rust
/// use cudarc::driver::{profiler_start, profiler_stop};
```

src/driver/safe.rs
```diff
@@ -1121,6 +1121,27 @@ impl CudaDeviceBuilder {
...
+///
+/// profiler_start()?;
+/// // Hotpath
+/// profiler_stop()?;
+/// // Now check your results
+/// // nsys profile -c cudaProfilerApi /path/to/bin
+/// // And this will profile only the hotpath.
```
coreylowman, line 1133: Nice inclusion 😀

PR #89, opened 2023-03-03
Updating launch config when less than 1024

coreylowman:
TODO: see if this is standard?


coreylowman:
Don't think pytorch does this

PR #104, opened 2023-03-24
Adding static linking potential.

Narsil:
When building binaries with cuda, the dependency on the libs is super large.

Biggest culprit is cublasLt which is 485Mb on its own.
Using static linking, shrinks everything into the binary which becomes ~130Mo (after stripping) in my case.
It also makes cublas loading much faster.

I chose to use an env variable for the type of linking since it avoids introducing new features and bloating the code with them.

Feel free to ignore if you don't deem this worthy !

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -20,15 +20,29 @@ fn link_cuda() {
     #[cfg(feature = "driver")]
     println!("cargo:rustc-link-lib=dylib=cuda");
     #[cfg(feature = "driver")]
-    println!("cargo:rustc-link-lib=dylib=cudart");
+    println!("cargo:rustc-link-lib=static=cudart_static");
     #[cfg(feature = "nvrtc")]
     println!("cargo:rustc-link-lib=dylib=nvrtc");
     #[cfg(feature = "curand")]
     println!("cargo:rustc-link-lib=dylib=curand");
-    #[cfg(feature = "cublas")]
-    println!("cargo:rustc-link-lib=dylib=cublas");
-    #[cfg(feature = "cublas")]
-    println!("cargo:rustc-link-lib=dylib=cublasLt");
+
+    println!("cargo:rerun-if-env-changed=STATIC");
+    let is_static = std::env::var("STATIC").unwrap_or_else(|_| "0".to_string()) == "1";
+
+    if is_static {
+        #[cfg(feature = "cublas")]
+        println!("cargo:rustc-link-lib=dylib=stdc++");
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -20,15 +20,29 @@ fn link_cuda() {
...
-    println!("cargo:rustc-link-lib=dylib=cublas");
-    #[cfg(feature = "cublas")]
-    println!("cargo:rustc-link-lib=dylib=cublasLt");
+
+    println!("cargo:rerun-if-env-changed=STATIC");
+    let is_static = std::env::var("STATIC").unwrap_or_else(|_| "0".to_string()) == "1";
+
+    if is_static {
+        #[cfg(feature = "cublas")]
+        println!("cargo:rustc-link-lib=dylib=stdc++");
```
coreylowman, line 34: Do we need this one?

build.rs
```diff
@@ -20,15 +20,29 @@ fn link_cuda() {
     #[cfg(feature = "driver")]
     println!("cargo:rustc-link-lib=dylib=cuda");
     #[cfg(feature = "driver")]
-    println!("cargo:rustc-link-lib=dylib=cudart");
+    println!("cargo:rustc-link-lib=static=cudart_static");
```
coreylowman, line 23: Didn't realize until this PR, but we can actually remove this lib all together - we don't use any runtime apis

build.rs
```diff
@@ -20,15 +20,29 @@ fn link_cuda() {
...
     println!("cargo:rustc-link-lib=dylib=nvrtc");
     #[cfg(feature = "curand")]
     println!("cargo:rustc-link-lib=dylib=curand");
-    #[cfg(feature = "cublas")]
-    println!("cargo:rustc-link-lib=dylib=cublas");
-    #[cfg(feature = "cublas")]
-    println!("cargo:rustc-link-lib=dylib=cublasLt");
+
+    println!("cargo:rerun-if-env-changed=STATIC");
+    let is_static = std::env::var("STATIC").unwrap_or_else(|_| "0".to_string()) == "1";
```
coreylowman, line 30: Let's go with feature flag, it's more reproducible than environment variable and I think fits better within rust ecosystem at least.

coreylowman (changes requested):
I like this change, definitely want to add it in!

Narsil:
Superseeded by https://github.com/coreylowman/cudarc/pull/105 (allows me to keep my main for experimentation).

PR #105, opened 2023-03-25
Static linking (with feature flag).

Narsil:
When building binaries with cuda, the dependency on the libs is super large.

Biggest culprit is cublasLt which is 485Mb on its own.
Using static linking, shrinks everything into the binary which becomes ~130Mo (after stripping) in my case.
It also makes cublas loading much faster.


coreylowman:
Looks good, thanks for the PR!

PR #106, opened 2023-03-25
Adding opt-out sync alloc for older cards.

Narsil:
currently have an older GPU (GTX 970). With compute cap 5.2.

Those cards get Operation not supported for the alloc/free async part.

This PR should allow for this crate to work for such older cards by using sync alloc without stream.
Since it's sync and a superset it's fine. 
Since it's for old cards only (it seems) the feature is enabled by default.

Wdyt ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ keywords = [
 features = ["ci-check", "f16"]
 
 [features]
-default = ["std", "driver", "nvrtc", "cublas", "curand"]
+default = ["std", "driver", "nvrtc", "cublas", "curand", "alloc_async"]
 nvrtc = []
 driver = ["nvrtc"]
 cublas = []
@@ -34,6 +34,7 @@ std = []
 no-std = ["no-std-compat/std", "dep:spin"]
 f16 = ["dep:half"]
 ci-check = []
+alloc_async = []
 
 [dependencies]
 spin = { version = "0.9.4", optional = true, features = ["rwlock"], default-features = false }
--- a/src/driver/result.rs
+++ b/src/driver/result.rs
@@ -281,7 +281,11 @@ pub unsafe fn malloc_async(
     num_bytes: usize,
 ) -> Result<sys::CUdeviceptr, DriverError> {
     let mut dev_ptr = MaybeUninit::uninit();
+    #[cfg(feature = "alloc_async")]
     sys::cuMemAllocAsync(dev_ptr.as_mut_ptr(), num_bytes, stream).result()?;
+
+    #[cfg(not(feature = "alloc_async"))]
+    sys::cuMemAlloc_v2(dev_ptr.as_mut_ptr(), num_bytes).result()?;
     Ok(dev_ptr.assume_init())
 }
 
@@ -294,7 +298,15 @@ pub unsafe fn malloc_async(
 /// 2. The memory should have been allocated on this stream.
 /// 3. The memory should not have been freed already (double free)
 pub unsafe fn free_async(dptr: sys::CUdeviceptr, stream: sys::CUstream) -> Result<(), DriverError> {
-    sys::cuMemFreeAsync(dptr, stream).result()
+    #[cfg(feature = "alloc_async")]
+    {
+        sys::cuMemFreeAsync(dptr, stream).result()
+    }
+
+    #[cfg(not(feature = "alloc_async"))]
+    {
+        sys::cuMemFree_v2(dptr).result()
+    }
 }
 
 /// Sets device memory with stream ordered semantics.


Cargo.toml
```diff
@@ -34,6 +34,7 @@ std = []
 no-std = ["no-std-compat/std", "dep:spin"]
 f16 = ["dep:half"]
 ci-check = []
+alloc_async = []
```
coreylowman, line 37: Is it possible to check this at runtime instead of feature flag?

src/driver/result.rs
```diff
@@ -281,7 +281,11 @@ pub unsafe fn malloc_async(
     num_bytes: usize,
 ) -> Result<sys::CUdeviceptr, DriverError> {
     let mut dev_ptr = MaybeUninit::uninit();
+    #[cfg(feature = "alloc_async")]
     sys::cuMemAllocAsync(dev_ptr.as_mut_ptr(), num_bytes, stream).result()?;
+
+    #[cfg(not(feature = "alloc_async"))]
+    sys::cuMemAlloc_v2(dev_ptr.as_mut_ptr(), num_bytes).result()?;
```
coreylowman, line 288: Hmm we may have to do more to support this - I'm concerned this might interact poorly with multiple non-blocking streams. The current safe api is built around the assumption of stream ordered semantics, so this could lead to subtle timing bugs

Cargo.toml
Narsil: I'm not sure, the error is "Operation not supported". I'm not sure trying alloc and retrying with sync alloc is a good general approach. 

To run on this older card, I definitely need this bit of code, whether it's worth keeping in the library is not is not really clear.
Tbh, if I was maintaining, I wouldn't add this flag to the main library. I'm happy to keep this as a fork, until more demand and or clearer path for support.

src/driver/result.rs
Narsil: This is a synced global operation iiurc, meaning it will wait for all streams to finish + block alloc.

I'm far from 100% confident but the code I wrote with this seems to work correctly. And the semantics for cuda library seem to imply that blockign calls will both wait on all streams + sync op, meaning it's safe ops. It's my understanding at least.

coreylowman: I think since all allocs/copies for safe api go on default stream this should be fine. The one place where it may be messed up is using CudaStream which currently creates a non blocking stream https://github.com/coreylowman/cudarc/blob/main/src/driver/safe/core.rs#L226.

Probably if we detect this feature we should create Default stream kind https://github.com/coreylowman/cudarc/blob/main/src/driver/result.rs#L196

Narsil: But `cuMemAlloc_v2` doesn't specify the stream.

My understanding is that it should be behind all streams. I don't see any reference in the docs specifying anything about streams though. Maybe I'm wrong, and if not specified it just means send to default stream. 
Not sure how I could test/confirm this though.

coreylowman: Yeah allocation should be fine but other calls (like invoking kernels) could still use other streams. That's where the CudaStream & non blocking would come into play

Narsil: Also since this call is *sync*, the other cuda stream calls, would only cause issues if called from other threads, correct ?

coreylowman: Oh wait you mean cuMemAlloc_v2 is sync w.r.t the host? I'll have to double check. If so then yes. If it's still async wrt the host, then you could cause issues even when doing threaded

Narsil: I *think* it's sync. It's not explicitly mentionned in the doc I think, but since all async calls are explicitly marked as async, I assume.

coreylowman, line 288: Let's change the feature name to `sync_alloc`, reverse and remove it from the default features list. Would prefer opt-in to the non-standard behavior, rather than opting out of it.

```suggestion
    #[cfg(not(feature = "sync_alloc"))]
    sys::cuMemAllocAsync(dev_ptr.as_mut_ptr(), num_bytes, stream).result()?;

    #[cfg(feature = "sync_alloc")]
    sys::cuMemAlloc_v2(dev_ptr.as_mut_ptr(), num_bytes).result()?;
```

Narsil: That's a feature that removes something, no ?

I'll take it if you want to include in the lib.

Actually thinking about it, what about, not having a feature flag, and having 2 functions (pushing the burden on the user to figure it out, which IMO is better here, since it should be my choice to figure out which function to calll)

coreylowman: Yeah that could work - we'd have to figure out how to integrate into CudaDevice allocation functions I think

Narsil: Probably.

coreylowman:
Closing since #174  takes care of this

PR #109, opened 2023-03-27
Fix RangeHelper::bounds's handling of zero-length ranges

nkoppel:
Prevents `bounds` returning `None` when given a range of length zero by returning an exclusive end of each range. This comes with the minor limitation that this function can no longer handle `usize::MAX`, but revolves incorrect behavior in dfdx when concatenating to a tensor of size 0.


M1ngXU:
just as a note, ranges with negative lengths (1..0) will return a length of 0

nkoppel:
Yeah, `1..=0` is interpreted as a zero length range at 1, equivalent to `1..1`. This should probably produce None, but it doesn't seem like a huge deal to me.

nkoppel:
Ok, this behavior should now be fixed.

M1ngXU:
> Yeah, `1..=0` is interpreted as a zero length range at 1, equivalent to `1..1`. This should probably produce None, but it doesn't seem like a huge deal to me.

yeah probably, though just wanted to be sure that no weird behaviour would happen

PR #112, opened 2023-03-29
Adds CudaDevice::bind_to_thread and example of multi-threaded use

### LINKED ISSUE #108: Issue with thread safety intrinsics of `Arc<CudaDevice>`
I am not sure wrapping the CudaDevice within an `Arc` is living to it's full potential.

Here is a simple crate showcasing the issue: 

https://github.com/Narsil/axum_cudarc

It seems that it's an `Arc` the device can be shared, however, the allocation doesn't work meaning that the cuda ctx is not really thread safe.
Seems like a gotcha to me, and we shouldn't be able to do that. 
I'm still not sure what's the best way to include GPU code on a webserver (single thread owning the GPU and communication with CPU threads on the webserver, vs all webservers threads accessing the GPU and reusing 1 stream for ordering (if wanted).

In any case, it seems that the current behavior is not desirable.

Did I miss something ?

coreylowman:
Resolves #108

### DIFF
--- a/examples/06-threading.rs
+++ b/examples/06-threading.rs
@@ -1,0 +1,28 @@
+use cudarc::driver::{CudaDevice, DeviceSlice, DriverError};
+
+use std::thread;
+
+fn main() -> Result<(), DriverError> {
+    let dev = CudaDevice::new(0)?;
+
+    let dev1 = dev.clone();
+    let thread1 = thread::spawn(move || {
+        // You'll need to call bind_to_thread on each thread that the
+        // device is used on.
+        dev1.bind_to_thread()?;
+        dev1.alloc_zeros::<f32>(100)
+    });
+
+    let dev2 = dev;
+    let thread2 = thread::spawn(move || {
+        // Note we have the same call here.
+        dev2.bind_to_thread()?;
+        dev2.alloc_zeros::<f32>(100)
+    });
+
+    let a = thread1.join().unwrap()?;
+    let b = thread2.join().unwrap()?;
+    assert_eq!(a.len(), b.len());
+
+    Ok(())
+}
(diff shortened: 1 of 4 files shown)


examples/06-threading.rs
```diff
@@ -0,0 +1,28 @@
...
+
+fn main() -> Result<(), DriverError> {
+    let dev = CudaDevice::new(0)?;
+
+    let dev1 = dev.clone();
+    let thread1 = thread::spawn(move || {
+        // You'll need to call bind_to_thread on each thread that the
+        // device is used on.
+        dev1.bind_to_thread()?;
+        dev1.alloc_zeros::<f32>(100)
```
coreylowman, line 13: Should launch a kernel here that prints out a message to show that things are sequentially executed.

coreylowman:
Add another example `07-threaded-multi-device` that shows creating a device per thread

coreylowman:
> what happens if bind_to_thread is not called? ~UB~?
> 
> edit: well, then it should be ok

You can see what happens in the unit test I added, basically a DriverError(INVALID_CONTEXT_ERROR) is returned

src/driver/safe/threading.rs
```diff
@@ -0,0 +1,51 @@
...
+            dev2.bind_to_thread()?;
+            dev2.alloc_zeros::<f32>(10)
+        });
+
+        let _: crate::driver::CudaSlice<f32> = thread1.join().unwrap().unwrap();
+        let _: crate::driver::CudaSlice<f32> = thread2.join().unwrap().unwrap();
+    }
+
+    #[test]
+    #[should_panic = "DriverError(CUDA_ERROR_INVALID_CONTEXT"]
```
coreylowman, line 40: @M1ngXU

M1ngXU: just saw this

PR #125, opened 2023-04-10
Adds CudaDevice::null method & CudaSlice::device method

### LINKED ISSUE #124: Give access to underlying device from CudaSlice
Related to #122 

When creating CudaSlice from CUdeviceptr, we might have another CudaSlice instead of a CudaDevice. Should be able to reuse the device from the slice.

coreylowman:
Resolves #124

### DIFF
--- a/src/driver/safe/alloc.rs
+++ b/src/driver/safe/alloc.rs
@@ -81,6 +81,17 @@ impl<T> CudaSlice<T> {
         std::mem::forget(self);
         ptr
     }
+
+    /// Empties this mutable object (allocates a new )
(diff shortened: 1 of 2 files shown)


src/driver/safe/alloc.rs
```diff
@@ -81,6 +81,17 @@ impl<T> CudaSlice<T> {
         std::mem::forget(self);
         ptr
     }
+
+    /// Empties this mutable object (allocates a new )
+    pub fn replace_with_empty(&mut self) -> Self {
```
coreylowman, line 86: Should this be unsafe?

PR #148, opened 2023-05-11
rust kernel & example

ariasanovsky:
Run example with

```powershell
cargo run --release --example 08-rust-kernel
```

RustPtx builds ptx from a valid lib.rs with Command and ABI_PTX (experimental, expect bugs and old PTX versions). lib.rs requires a Cargo.toml.

### DIFF
--- a/examples/08-rust-kernel.rs
+++ b/examples/08-rust-kernel.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::{PathBuf, Path};
 
 use cudarc::{
     driver::{CudaDevice, DriverError, LaunchAsync, LaunchConfig},
@@ -8,17 +8,13 @@ use cudarc::{
 fn main() -> Result<(), DriverError> {
     let dev = CudaDevice::new(0)?;
 
-    let kernel: PathBuf = "examples/rust-kernel/src/lib.rs".into();
-    let rust_ptx: Result<PtxCrate, _> = kernel.try_into();
-    
-    let mut rust_ptx = rust_ptx.unwrap();
-    
-    rust_ptx.build_ptx().unwrap();
-    
-    let kernels = rust_ptx.ptx_files().unwrap();
+    // use compile_crate_to_ptx to build and rust kernels in pure rust
+    // uses experimental ABI_PTX
+    let kernel_path: PathBuf = "examples/rust-kernel/src/lib.rs".into();
+    let kernels = PtxCrate::compile_crate_to_ptx(&kernel_path).unwrap();
     let kernel = kernels.first().unwrap();
 
-    // You can load a function from a pre-compiled PTX like so:
+    // load the ptx file...
     println!("loading...");
     dev.load_ptx(kernel.clone(), "kernel", &["kernel"])?;
     println!("loaded!");
@@ -39,12 +35,14 @@ fn main() -> Result<(), DriverError> {
     let b_host = dev.sync_reclaim(b_dev.clone())?;
 
     println!("a_host {a_host:?}");
-    
     println!("b_host {b_host:?}");
-    
     println!("a_host_2 {a_host_2:?}");
-
-    println!("cleaned successfully? {:?}", rust_ptx.clean());
     
+    // we can also manage and clean up the build ptx files with a PtxCrate
+    let mut rust_ptx: PtxCrate = kernel_path.try_into().unwrap();
+    rust_ptx.build_ptx().unwrap();
+    let _kernel = rust_ptx.ptx_files().unwrap().first().unwrap();
+    println!("cleaned successfully? {:?}", rust_ptx.clean());
+
     Ok(())
 }
--- a/examples/08-rust-kernel.rs
+++ b/examples/08-rust-kernel.rs
@@ -1,43 +1,69 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, array};
 
 use cudarc::{
     driver::{CudaDevice, DriverError, LaunchAsync, LaunchConfig},
     nvrtc::PtxCrate,
 };
 
+use std::time::Instant;
+
 fn main() -> Result<(), DriverError> {
     let dev = CudaDevice::new(0)?;
 
-    // use compile_crate_to_ptx to build and rust kernels in pure rust
+    // use compile_crate_to_ptx to build kernels in pure Rust
     // uses experimental ABI_PTX
     let kernel_path: PathBuf = "examples/rust-kernel/src/lib.rs".into();
     let kernels = PtxCrate::compile_crate_to_ptx(&kernel_path).unwrap();
     let kernel = kernels.first().unwrap();
 
     // load the ptx file...
-    println!("loading...");
     dev.load_ptx(kernel.clone(), "rust_kernel", &["square_kernel"])?;
-    println!("loaded!");
-
+    
     // and then retrieve the function with `get_func`
     let f = dev.get_func("rust_kernel", "square_kernel").unwrap();
 
-    let a_host = [1.0f32, 2., 3.];
+    const GRID: u32 = 32;
+    const BLOCK: u32 = 512;
+    const THREADS: u32 = GRID * BLOCK;
+
+    // https://developer.nvidia.com/blog/cuda-pro-tip-occupancy-api-simplifies-launch-configuration/
+    let cfg = LaunchConfig {
+        grid_dim: (GRID, 1, 1),
+        block_dim: (BLOCK, 1, 1),
+        shared_mem_bytes: 0,
+    };
+    // or use let cfg = LaunchConfig::for_num_elems(n);
+
+    let a_host: [f32; THREADS as usize] = array::from_fn(|i| {
+        i as f32
+    });
 
     let a_dev = dev.htod_copy(a_host.into())?;
     let mut b_dev = a_dev.clone();
 
-    let n = 3;
-    let cfg = LaunchConfig::for_num_elems(n);
-    unsafe { f.launch(cfg, (&a_dev, &mut b_dev, n as i32)) }?;
-
-    let a_host_2 = dev.sync_reclaim(a_dev.clone())?;
-    let b_host = dev.sync_reclaim(b_dev.clone())?;
-
-    println!("a_host {a_host:?}");
-    println!("b_host {b_host:?}");
-    println!("a_host_2 {a_host_2:?}");
+    println!("squaring 0.0, ..., {}.0 with cuda kernels...", THREADS - 1);
     
+    let start = Instant::now();
+    unsafe { f.launch(cfg, (&a_dev, &mut b_dev, THREADS as i32)) }?;
+    let duration = start.elapsed();
+    println!("launch duration: {:?}", duration);
+    
+    let start = Instant::now();
+    let b_host = dev.sync_reclaim(b_dev.clone())?;
+    let duration = start.elapsed();
+    println!("duration to get results: {:?}", duration);
+    
+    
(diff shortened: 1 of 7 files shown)


src/nvrtc/safe.rs
```diff
@@ -222,6 +222,51 @@ impl CompileOptions {
...
+    kernel_path: PathBuf,
+}
+
+use std::process::Command;
+impl RustPtx {
+    pub fn new(kernel_path: PathBuf) -> RustPtx {
+        RustPtx { kernel_path }
+    }
+
+    pub fn build_ptx(&self) -> Result<PathBuf, String> {
```
coreylowman, line 235: What are your thoughts on turning this into a single function that returns a Ptx object? I'm thinking interface like this:
```rust
pub fn compile_rust_to_ptx<S: AsRef<Path>>(kernel_path: S) -> Ptx {
    ... //current code
    Ptx::from_file(ptx_path)
}
```

coreylowman (commented):
Very cool! I really like the idea of having a crate of kernels

examples/rust-kernel/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+name = "kernel"
+path = "src/lib.rs"
+build = "cargo +nightly rustc --lib --target nvptx64-nvidia-cuda --release -- --emit asm"
+crate-type = ["cdylib"]
```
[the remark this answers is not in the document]
ariasanovsky: crate-type seemed necessary, but maybe it can be circumvented. I left build in as a comment for the curious

src/nvrtc/safe.rs
ariasanovsky: yep, just finished up a PtxCrate struct that has a TryFrom<PathBuf> and methods ptx_files, clean, and build_ptx

examples/08-rust-kernel.rs
```diff
@@ -1,17 +1,26 @@
...
+    let mut rust_ptx = rust_ptx.unwrap();
+    
     rust_ptx.build_ptx().unwrap();
+    
+    let kernels = rust_ptx.ptx_files().unwrap();
+    let kernel = kernels.first().unwrap();
 
     // You can load a function from a pre-compiled PTX like so:
     println!("loading...");
-    dev.load_ptx(Ptx::from_file("examples/rust-kernel/target/nvptx64-nvidia-cuda/release/kernel.ptx"), "kernel", &["kernel"])?;
```
ariasanovsky: PtxCrate now manages its ptx files in a Vec. Much less bloat.

examples/08-rust-kernel.rs
```diff
@@ -34,6 +43,8 @@ fn main() -> Result<(), DriverError> {
     println!("b_host {b_host:?}");
     
     println!("a_host_2 {a_host_2:?}");
+
+    println!("cleaned successfully? {:?}", rust_ptx.clean());
```
ariasanovsky, line 47: PtxCrate also cleans up after itself if you ask it to. Potentially this could be added to the destructor but for now it's opt-in in case users want to keep their residual ptx.

examples/rust-kernel/src/lib.rs
```diff
@@ -1,4 +1,3 @@
-#![cfg(feature = "nightly")]// nightly only due to experimental features (also improves linting linter)
```
ariasanovsky: This caused a bug with the cargo command I was using.

examples/rust-kernel/Cargo.toml
```diff
@@ -8,5 +8,5 @@ edition = "2021"
 [lib]
 name = "kernel"
 path = "src/lib.rs"
-build = "cargo +nightly rustc --lib --target nvptx64-nvidia-cuda --release -- --emit asm"
+# build = "cargo +nightly rustc --lib --target nvptx64-nvidia-cuda --release -- --emit asm"
```
ariasanovsky, line 11: Leaving this in a comment for convenience/curiosity.

src/nvrtc/safe.rs
```diff
@@ -222,47 +223,147 @@ impl CompileOptions {
     }
 }
 
-pub struct RustPtx {
-    kernel_path: PathBuf,
+#[derive(Debug)]
```
ariasanovsky, line 226: RustPtx -> PtxCrate and PtxCrateKind (planned to eventually support standalone compilation directly through rustc. Standalone is possible with

```rust
        let output = Command::new("rustc")
            .arg("+nightly")
            .arg("--crate-type=cdylib")
            .arg("--target")
            .arg("nvptx64-nvidia-cuda")
            .arg(kernel_path) // change the name
            .arg("--")
            .arg("--emit")
            .arg("asm")
            .output()
            .map_err(|e| format!("Failed to execute command: {}", e))?;
```
or some variant of that. It produces massive ptx files. I couldn't figure out how to add the flags -C and opt-level=3

ariasanovsky (commented):
PtxCrate struct hopefully keeps things cleaner

src/nvrtc/safe.rs
ariasanovsky: Other idea: create a temporary Cargo.toml to accompany a standalone kernel lib.rs.

src/nvrtc/safe.rs
ariasanovsky: PtxCrate::compile_crate_to_ptx is now a-go

examples/08-rust-kernel.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use std::time::Instant;
+
+fn main() -> Result<(), DriverError> {
+    let dev = CudaDevice::new(0)?;
+
+    // use compile_crate_to_ptx to build kernels in pure Rust
+    // uses experimental ABI_PTX
+    let kernel_path: PathBuf = "examples/rust-kernel/src/lib.rs".into();
+    let kernels = PtxCrate::compile_crate_to_ptx(&kernel_path).unwrap();
```
coreylowman, line 16: Can you include the type of this variable? Just so its more clear what the type is 😄

src/nvrtc/safe.rs
coreylowman: Hmm yeah, I can't quite tell which version developers of kernels would prefer. My instinct is to just choose whichever one is simpler so we aren't over complicated things. Since you've already implemented the cargo variant, maybe we just remove the standalone for now? It will be okay to introduce breaking changes here in the future, since this depends on some very not stable features. It'd also be easier to develop & test if we just have 1 version

Thoughts?

src/nvrtc/safe.rs
```diff
@@ -222,6 +224,180 @@ impl CompileOptions {
...
+    pub fn compile_crate_to_ptx<S: AsRef<Path>>(kernel_path: S) -> Result<Vec<Ptx>, String> {
+        let kernel_path: PathBuf = kernel_path.as_ref().into();
+        let mut rust_ptx: Self = kernel_path.try_into()?;
+        rust_ptx.build_ptx()?;
+        Ok(rust_ptx.take_kernels().unwrap())
+    }
+
+    pub fn take_kernels(self) -> Option<Vec<Ptx>> {
+        self.kernels
+    }
```
coreylowman, line 316: I think this has *potential* to have weird interactions with `clean()`. If someone calls .take_kernels() and then calls .clean(), the kernels would be invalidated, but the ptx objects would still be exist. Ptx::from_file just stores the file path, and only reads it on use.

We might want to separately update Ptx::from_file to read the contents on creation, or just remove this method for now.

I like this idea though, and I also like `clean`. I was actually thinking that we could do `impl Drop for PtxCrate` and call `clean`

coreylowman (commented):
Just some small comments. Will need to add docstrings, but I can help with that!

src/nvrtc/safe.rs
ariasanovsky: Sure, that's cool. Also I like what you did with the command arguments for the C/C++ -> PTX emitter. Definitely a good candidate for future breaking changes to give more features once this has time to settle.

src/nvrtc/safe.rs
ariasanovsky: My thoughts were that if someone wants to take ownership of the kernels without a PtxCrate, it's on them to clean if they want. I hesitate to do `impl Drop for PtxCrate` because I fear it would interfere with `build.rs` scripts that create and then drop a `PtxCrate`.

examples/08-rust-kernel.rs
ariasanovsky: Added docstrings for everything.

ariasanovsky:
Le me know also if you can `cargo run --release --example 08-rust-kernel`. Would definitely want this working on at least 2 machines. I have it set to $16384 = 256 \times 64$ threads.

src/nvrtc/safe.rs
coreylowman: Yeah let's just remove it entirely 😄

src/nvrtc/safe.rs
coreylowman: +1 to making it a module level free function.

src/nvrtc/safe.rs
```diff
@@ -222,6 +225,257 @@ impl CompileOptions {
...
+
+                if output.status.success() {
+                    let ptx_path = project_dir.join("target/nvptx64-nvidia-cuda/release");
+                    if ptx_path.exists() {
+                        let ptx_files: Vec<Ptx> = fs::read_dir(ptx_path)
+                        .map_err(|e| e.to_string())?
+                        .filter_map(|entry| {
+                            let path = entry.unwrap().path();
+                            if let Some("ptx") = path.extension().and_then(|ext| ext.to_str()) {
+                                Some(Ptx::from_file(path))
```
coreylowman, line 453: Let's change this to
1. Read file contents using std::fs
2. use `Ptx::from_src` with the contents
Then we don't have to worry about files being removed invalidating the Ptx objects.

src/nvrtc/safe.rs
```diff
@@ -222,6 +225,257 @@ impl CompileOptions {
...
+impl TryFrom<PathBuf> for PtxCrate {
+    type Error = String;
+
+    fn try_from(value: PathBuf) -> Result<Self, Self::Error> {
+        let project = value.try_into()?;
+        Ok(PtxCrate { project, kernels: None })
+    }
+}
+
+/// `PtxCrate` provides methods to compile a Rust crate to CUDA PTX code,
```
coreylowman, line 306: Nice doc strings!! These are really helpful

BTW I think you can probably move this to on top of the struct definition instead of the impl block, that's generally the way to do it 😄

ariasanovsky: Hopefully this is better now. I can't `cargo doc` right now because

```bash
  --> src/lib.rs:20:20
   |
20 | //! | cudnn | - | [cudnn::result] | [cudnn::sys] |
   |                    ^^^^^^^^^^^^^ no item named `cudnn` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: unresolved link to `cudnn::sys`
  --> src/lib.rs:20:38
   |
20 | //! | cudnn | - | [cudnn::result] | [cudnn::sys] |
   |                                      ^^^^^^^^^^ no item named `cudnn` in scope

warning: `cudarc` (lib doc) generated 2 warnings
```

ariasanovsky:
I don't like my ergonomic choices here and don't want feature cruft clogging a backend crate. I've moved my schemes to [spindle](https://crates.io/crates/spindle).

PR #150, opened 2023-05-16
WIP: Add occupancy to driver::result

tthebst:
Some questions: 

How do you plan to return multiple values (cuOccupancyMaxPotentialBlockSize)? 

How would the safe API look like?


coreylowman (approved):
I think this looks good. Happy to have the safe API be a separate PR.

I think the safe API would involve adding methods to the `CudaFunction` struct. It seems like most of the safety involves having a function that exists, which the safe api already covers.

tthebst:
Okay. Let's merge this.

coreylowman:
Sweet, thanks for this contribution!

PR #152, opened 2023-05-18
Add functions to interact with external memory

ViliamVadocz:
- Add unsafe functions to `crate::driver::result` in a new `external_memory` module which wrap raw bindings for interacting with external memory.
- Add safe(?) wrapper in the form of two new types `ExternalMemory` and `MappedBuffer` which abstract away interactions with external memory and to manage cleanup on drop.

I did not add any tests because I don't yet know how to get a valid file descriptor using only CUDA. At least I confirmed that this interface works with a file descriptor exported from Vulkan (via `vulkano` crate). The code exists in my repo https://github.com/ViliamVadocz/nvidia-video-codec-sdk on the `cudarc-external-memory` branch (very experimental and unorganized).

### DIFF
--- a/src/driver/result.rs
+++ b/src/driver/result.rs
@@ -574,6 +574,12 @@ pub mod external_memory {
 
     use super::{sys, DriverError};
 
+    /// Imports an external memory object, in this case an OpaqueFd.
+    ///
+    /// See [cuda docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html#group__CUDA__EXTRES__INTEROP_1g52aba3a7f780157d8ba12972b2481735)
+    ///
+    /// # Safety
+    /// `size` must be the size of the size of the memory object in bytes.
     #[cfg(unix)]
     pub unsafe fn import_external_memory_opaque_fd(
         fd: std::os::fd::RawFd,
@@ -586,11 +592,16 @@ pub mod external_memory {
             size,
             ..Default::default()
         };
-        unsafe { sys::cuImportExternalMemory(external_memory.as_mut_ptr(), &handle_description) }
-            .result()?;
+        sys::cuImportExternalMemory(external_memory.as_mut_ptr(), &handle_description).result()?;
         Ok(external_memory.assume_init())
     }
 
+    /// Imports an external memory object, in this case an OpaqueWin32 handle.
+    ///
+    /// See [cuda docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html#group__CUDA__EXTRES__INTEROP_1g52aba3a7f780157d8ba12972b2481735)
+    ///
+    /// # Safety
+    /// `size` must be the size of the size of the memory object in bytes.
     #[cfg(windows)]
     pub unsafe fn import_external_memory_opaque_win32(
         handle: std::os::windows::io::RawHandle,
@@ -608,17 +619,29 @@ pub mod external_memory {
             size,
             ..Default::default()
         };
-        unsafe { sys::cuImportExternalMemory(external_memory.as_mut_ptr(), &handle_description) }
-            .result()?;
+        sys::cuImportExternalMemory(external_memory.as_mut_ptr(), &handle_description).result()?;
         Ok(external_memory.assume_init())
     }
 
+    /// Destroys an external memory object.
+    ///
+    /// See [cuda docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html#group__CUDA__EXTRES__INTEROP_1g1b586dda86565617e7e0883b956c7052)
+    ///
+    /// # Safety
+    /// - Any mapped buffers onto this object must already be freed.
+    /// - The external memory must only be destroyed once.
     pub unsafe fn destroy_external_memory(
         external_memory: sys::CUexternalMemory,
     ) -> Result<(), DriverError> {
-        unsafe { sys::cuDestroyExternalMemory(external_memory) }.result()
+        sys::cuDestroyExternalMemory(external_memory).result()
     }
 
+    /// Maps a buffer onto an imported memory object.
+    ///
+    /// See [cuda docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html#group__CUDA__EXTRES__INTEROP_1gb9fec33920400c70961b4e33d838da91)
+    ///
+    /// # Safety
+    /// Mapped buffers may overlap.
     pub unsafe fn get_mapped_buffer(
         external_memory: sys::CUexternalMemory,
         offset: u64,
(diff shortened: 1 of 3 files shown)


src/driver/safe/external_memory.rs
```diff
@@ -0,0 +1,124 @@
...
+impl CudaDevice {
+    /// Import external memory from a [`File`].
+    ///
+    /// # Safety
+    /// `size` must be the size of the external memory.
+    #[cfg(any(unix, windows))]
+    pub unsafe fn import_external_memory(
+        self: &Arc<Self>,
+        file: File,
+        size: u64,
```
coreylowman, line 18: Could you use `file`'s metadata for this? https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len

src/driver/result.rs
```diff
@@ -568,3 +568,108 @@ pub unsafe fn launch_kernel(
...
+    /// Frees device memory.
+    ///
+    /// see [cuda docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g89b3f154e17cc89b6eea277dbdf5c93a)
+    ///
+    /// # Safety
+    /// - Memory must only be freed once.
+    /// - All async accesses to this pointer must have been completed.
+    pub unsafe fn memory_free(device_ptr: sys::CUdeviceptr) -> Result<(), DriverError> {
+        sys::cuMemFree_v2(device_ptr).result()
+    }
```
coreylowman, line 674: We have free_async a top level mod, want to move this up a level? It doesn't seem to be specific to the external memory module.

src/driver/safe/external_memory.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// An abstraction for a mapped buffer for some external memory.
+///
+/// This struct can be created via [`ExternalMemory::map_range`] or [`ExternalMemory::map_all`].
+/// The underlying mapped buffer will be freed when this struct is dropped.
+#[derive(Debug)]
+pub struct MappedBuffer<'a> {
+    device_ptr: sys::CUdeviceptr,
+    len: usize,
+    _marker: PhantomData<&'a mut ExternalMemory>,
```
coreylowman, line 105: Thoughts on actually having a reference here? If we do this then in the drop implementation, we could use `result::free_async(self.device_ptr, self.external_memory.device.stream)`

src/driver/safe/external_memory.rs
```diff
@@ -0,0 +1,124 @@
...
+/// An abstraction for imported external memory.
+///
+/// This struct can be created via [`CudaDevice::import_external_memory`].
+/// The imported external memory will be destroyed when this struct is dropped.
+#[derive(Debug)]
+pub struct ExternalMemory {
+    external_memory: sys::CUexternalMemory,
+    size: u64,
+    _device: Arc<CudaDevice>,
+    _file: ManuallyDrop<File>,
```
coreylowman, line 48: I'm not familiar with ManuallyDrop, what's the benefit of wrapping this?

src/driver/safe/external_memory.rs
ViliamVadocz: I experimented with this, and the `File` objects I get [from Vulkan](https://docs.rs/vulkano/latest/vulkano/memory/struct.DeviceMemory.html#method.export_fd) do not fill the metadata. When I looked at the length it was 0.
I could just use the metadata and enforce that the caller sets it correctly, but that seemed more error-prone than asking for it explicitly like this.

src/driver/result.rs
ViliamVadocz: Makes sense. Are you ok with the name `memory_free`?

src/driver/safe/external_memory.rs
ViliamVadocz: [`ManuallyDrop`](https://doc.rust-lang.org/stable/std/mem/struct.ManuallyDrop.html) allows you to avoid calling `drop` of the inner type when it goes out of scope. Ideally I would just use [`std::mem::forget`](https://doc.rust-lang.org/std/mem/fn.forget.html) in the `Drop` implementation, but I can't because I cannot move an attribute out while I have a mutable reference. I also cannot do the usual trick of [`std::mem::take`](https://doc.rust-lang.org/std/mem/fn.take.html) because `File` does not implement `Default`.

The reason we need to avoid dropping is because on Unix, the CUDA driver takes ownership of the file handle and any operations on it after are UB. On Windows though, you have to drop the handle yourself. See the [cuda docs here](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html#group__CUDA__EXTRES__INTEROP_1g52aba3a7f780157d8ba12972b2481735) (scroll down until you get to "If [CUDA_EXTERNAL_MEMORY_HANDLE_DESC::type](https://docs.nvidia.com/cuda/cuda-driver-api/structCUDA__EXTERNAL__MEMORY__HANDLE__DESC__v1.html#structCUDA__EXTERNAL__MEMORY__HANDLE__DESC__v1_1d4e3663348d28278d066980b422ab70e) is [CU_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggbca0bdde9a33c47058b5c97f21e2edd864efb8402f2268e489336396b2048c07) ...").

If you prefer to avoid `ManuallyDrop` I could also use an `Option` and that would allow me to use `std::mem::take` and `std::mem::forget`.

src/driver/safe/external_memory.rs
ViliamVadocz: I am unsure if that is sound, because from the [CUDA docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXTRES__INTEROP.html#group__CUDA__EXTRES__INTEROP_1gb9fec33920400c70961b4e33d838da91) it explicitly mentions that `cuMemFree` must be used:

> The returned pointer devPtr must be freed using [cuMemFree](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g89b3f154e17cc89b6eea277dbdf5c93a).

src/driver/result.rs
coreylowman: Yep!

src/driver/safe/external_memory.rs
coreylowman: Oh great point, can you add a note in the code about that?

src/driver/safe/external_memory.rs
coreylowman: Oh weird, okay. Yeah I think that's fine then

src/driver/safe/external_memory.rs
coreylowman: Oh cool. I like ManuallyDrop for this, nice usage!

src/driver/safe/external_memory.rs
ViliamVadocz: I added a note to `get_mapped_buffer`.

coreylowman (approved):
Nice addition, thanks for the contribution 👍

PR #154, opened 2023-06-04
Adds an attribute method for device and a .rustfmt.toml

emchristiansen:
The .rustfmt.toml ensures formatting is respected when inside a Cargo workspace that might have different formatting.


coreylowman:
Thanks for the contribution 👍

PR #155, opened 2023-06-06
Improve external memory interface

ViliamVadocz:
While trying to use the API I added in #152, I ran into usability issues.
Originally I had `MappedBuffer` hold a `PhantomData<&mut ExternalMemory>` which does make sense in principle, since it is essentially like a view, but it made it difficult to return `MappedBuffer` from functions.
By taking ownership of `ExternalMemory`, the drop order is still preserved, while making the API a bit easier to use in practice.

It does mean that you cannot map an `ExternalMemory` multiple times (before this you could `drop` the `MappedBuffer` to unmap), but I do not need that for my use-case, and in the future something like `MappedBuffer::unmap() -> ExternalMemory` could be added if it is needed.

This is a breaking change because it removes lifetime parameters and changes the function signatures of the mapping functions. Hopefully this is not a problem since a new version of cudarc has not been published since my previous PR.

### DIFF
--- a/src/driver/safe/external_memory.rs
+++ b/src/driver/safe/external_memory.rs
@@ -1,4 +1,4 @@
-use core::{marker::PhantomData, mem::ManuallyDrop};
+use core::mem::ManuallyDrop;
 use std::fs::File;
 use std::ops::Range;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


coreylowman:
Looks good to me 👍

PR #162, opened 2023-07-04
Adding RAII version for the profiler.

Narsil:
In some context this might be easier to use than calling `profiler_stop()` explicitly.


coreylowman (approved):
Looks easy to use, thanks!

PR #146, opened 2023-05-09
NCCL integration

Thenerdstation:
Opening as a draft PR to get early feedback that I am doing the bindgen correctly. I am currently still working on the result and safe wrappings. 

Fixes #129

PR #164, opened 2023-07-04
NCCL support

Narsil:
- Fixed a few stuff from original PR
- created a very shallow thread rank initializer

Linked to https://github.com/coreylowman/cudarc/pull/146


src/driver/safe/alloc.rs
```diff
@@ -302,7 +302,10 @@ impl CudaDevice {
         dst: &mut [T],
     ) -> Result<(), result::DriverError> {
         assert_eq!(src.len(), dst.len());
-        unsafe { result::memcpy_dtoh_async(dst, *src.device_ptr(), self.stream) }?;
+        unsafe {
+            result::ctx::set_current(self.cu_primary_ctx)?;
```
coreylowman, line 306: Related to #165

src/nccl/result.rs
```diff
@@ -0,0 +1,343 @@
...
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         write!(f, "{self:?}")
+//     }
+// }
+
+pub unsafe fn comm_finalize(comm: sys::ncclComm_t) -> Result<NcclStatus, NcclError> {
+    unsafe { sys::ncclCommFinalize(comm).result() }
+}
+
+pub unsafe fn comm_destry(comm: sys::ncclComm_t) -> Result<NcclStatus, NcclError> {
```
coreylowman, line 48:
```suggestion
pub unsafe fn comm_destroy(comm: sys::ncclComm_t) -> Result<NcclStatus, NcclError> {
```

src/nccl/result.rs
```diff
@@ -0,0 +1,343 @@
...
+    #[test]
+    fn single_thread() {
+        let n_devices = CudaDevice::count().unwrap() as usize;
+        let n = 2;
+
+        let mut devs = vec![];
+        let mut sendslices = vec![];
+        let mut recvslices = vec![];
+        for i in 0..n_devices {
+            let dev = CudaDevice::new(i).unwrap();
```
coreylowman, line 267: If you create the new device outside the loop can you call bind_to_thread() in this loop and below in the second loop?

src/nccl/safe.rs
```diff
@@ -0,0 +1,271 @@
...
+    #[test]
+    fn test_all_reduce() {
+        let mut threads = vec![];
+        let n = 2;
+        let n_devices = CudaDevice::count().unwrap() as usize;
+        let id = Id::new().unwrap();
+        for i in 0..n_devices {
+            threads.push(std::thread::spawn(move || {
+                let dev = CudaDevice::new(i).unwrap();
+                let comm = Comm::from_rank(dev.clone(), n_devices, id).unwrap();
```
coreylowman, line 256: I'm wondering if we should force creationg Comm's via some sort of `Vec<CudaDevice>` to avoid errors with id/n_devices. Something like:

```rust
fn init_group(devices: &[Arc<CudaDevice>]) -> Vec<Comm> { ... }

let devs: Vec<CudaDevice> = ...;
let comms: Vec<Comm> = init_group(&devs);
for comm in comms.drain(..) {
    std::thread::spawn(move || { ... }
}
```

Narsil: The group must be initialized from different threads (the call is blocking).
We could do your solution with `group_start/group_end`, but I'm not sure how if sending the comm is actually thread safe (in addition with the cuda context issue).

This is not exactly clear from the docs: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcomminitrank

vs 

https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#c.ncclCommInitAll

and multiple groups might be initialized. (for the id).

From rank might also be called by different processes. 

If you're Ok I'll try to enable rusty versions of all those various calls (masking groupStart).

src/nccl/safe.rs
```diff
@@ -0,0 +1,348 @@
...
+                    .try_into()
+                    .expect("World_size cannot be casted to i32"),
+                id.id,
+                rank.try_into().expect("Rank cannot be cast to i32"),
+            )?;
+            comm.assume_init()
+        };
+        Ok(Self { comm, device })
+    }
+}
```
Narsil, line 166: This is the core of the API for initiating Comm.

src/driver/safe/core.rs
```diff
@@ -176,7 +181,7 @@ impl Drop for CudaDevice {
 /// by the [CudaFunction]**.
 #[derive(Debug)]
 pub struct CudaSlice<T> {
-    pub(crate) cu_device_ptr: sys::CUdeviceptr,
+    pub cu_device_ptr: sys::CUdeviceptr,
```
coreylowman, line 184: This can be accessed through the DevicePtr traits

src/driver/safe/core.rs
```diff
@@ -34,7 +34,7 @@ pub struct CudaDevice {
     pub(crate) cu_device: sys::CUdevice,
     pub(crate) cu_primary_ctx: sys::CUcontext,
     /// The stream that all work is executed on.
-    pub(crate) stream: sys::CUstream,
+    pub stream: sys::CUstream,
```
coreylowman, line 37: This is already accessible through the cu_stream method

src/nccl/safe.rs
```diff
@@ -0,0 +1,348 @@
...
+define_nccl_type!(f32, sys::ncclDataType_t::ncclFloat32);
+define_nccl_type!(f64, sys::ncclDataType_t::ncclFloat64);
+define_nccl_type!(i8, sys::ncclDataType_t::ncclInt8);
+define_nccl_type!(i32, sys::ncclDataType_t::ncclInt32);
+define_nccl_type!(i64, sys::ncclDataType_t::ncclInt64);
+define_nccl_type!(u8, sys::ncclDataType_t::ncclUint8);
+define_nccl_type!(u32, sys::ncclDataType_t::ncclUint32);
+define_nccl_type!(u64, sys::ncclDataType_t::ncclUint64);
+define_nccl_type!(char, sys::ncclDataType_t::ncclUint8);
+
```
coreylowman, line 77:
```suggestion
#[cfg(feature = "f16")]
define_nccl_type!(half::f16, sys::ncclDataType_t:: ncclFloat16);
#[cfg(feature = "f16")]
define_nccl_type!(half::bf16, sys::ncclDataType_t:: ncclBfloat16);
```

coreylowman:
This looks good to me, just a couple of small changes and then I think we can merge 👍

Narsil:
Should be good now.

NCCL has a flag for BFloat16 support. I added it automatically (not sure it's an issue given the enum value is the last one, and the Dtype common is probably not really used).

Narsil:
Okay, after too much time investigating. It seems the Cuda Driver (not NCCL) is using a global MUTEX which makes multithread/multigpu quite useless.

https://forums.developer.nvidia.com/t/cuda-wont-concurrently-run-kernels-on-multiple-devices-from-within-same-process/240388
https://forums.developer.nvidia.com/t/multithreaded-tensorrt-performance-drops-dramatically/184882/8
https://forums.developer.nvidia.com/t/cuda-introduces-heavy-locks/61357

Basically all threads will contend for that lock not enabling the CPU to send kernels fast enough.

If there's any way to avoid that mutex as it makes the code significantly simpler to be multi thread rather than multiprocess.
But the tone in the related issues makes me think it's a known issue and nothing is going to be done about it.

coreylowman:
Oh. Wait that's a pretty big deal if its at the driver level, isn't it? For DDP for instance I was hoping we could use multi threading, but that would incur some serious slow downs due to locking. 🤔

Thenerdstation:
Thanks for finishing this! And sorry for the ghosting, life was getting in the way of open source 🙈

PR #167, opened 2023-07-07
Implement Gemm<bf16> for CudaBlas.

LaurentMazare:
(first thanks for this neat crate, it's very helpful)
This PR adds the `Gemm` trait for `bf16` so that cublas can easily be used on this type, the code is very similar to the `f16` variant but with the cuda types adapted to `bf16`.
The gemm `f16` test has also been adapted to test this variant too, all this being behind the `f16` feature flag.


coreylowman:
Awesome, thanks!

PR #174, opened 2023-08-10
Enabling runtime choice for mallocAsync or sync version.

Narsil:
Initial issue: https://github.com/huggingface/candle/issues/353

Basically cudarc fails on Windows even on recent cards, but also on older cards.

The core reason is that `cuMemAllocAsync` is not supported on those devices/OS.
This PR enables it by choosing between sync and async at runtime in the intended way by CUDA https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html  (Check supported platform).

Alternatives:

I would have preferred to use a compile time flag for this, since it seems always preferable to use async where available and sync elsewhere.

Previous work along that line.
https://github.com/coreylowman/cudarc/pull/106

I think we could have had a feature enabled in `build.rs` (https://github.com/rust-lang/cargo/issues/5499) for that particular thing.
However I was unable to find an easy way to detect that capacity using `nvidia-smi` (or any other expected CLI).
Meaning in order to do that at compile time I think we would have to compile first `cudaDeviceGetAttribute` small program in order to be able to call it within `build.rs`. This seems way overengineered and therefore I didn't do that.

I haven't measure performance with the if everywhere.

Another variant of this PR would be to keep track of wether the Slice was initialized with Async or not, to remove the if within `Drop`. (We could also store it directly on `CudaDevice` to remote the if within `alloc`.)
I deemed this as a performance feature and therefore didn't do it in this PR.


Narsil:
@coreylowman

coreylowman:
This looks fine to me, I like the runtime detection better than adding an additional compiler flag TBH.

@Narsil it looks like this deletes some examples, can you restore those? They are probably juts from out-dated fork

src/driver/safe/alloc.rs
```diff
@@ -107,10 +107,25 @@ impl CudaDevice {
 }
 
 impl CudaDevice {
+    pub fn is_async(self: &Arc<Self>) -> Result<bool, result::DriverError> {
+        let value = unsafe {
+            result::device::get_attribute(
+                self.cu_device,
+                sys::CUdevice_attribute_enum::CU_DEVICE_ATTRIBUTE_MEMORY_POOLS_SUPPORTED,
+            )?
```
coreylowman, line 115: Thoughts on storing this as an attribute of CudaDevice, so we only have to run it once when CudaDevice is created?

src/driver/safe/core.rs
```diff
@@ -194,7 +194,11 @@ impl<T> Drop for CudaSlice<T> {
     fn drop(&mut self) {
         self.device.bind_to_thread().unwrap();
         unsafe {
-            result::free_async(self.cu_device_ptr, self.device.stream).unwrap();
+            if self.device.is_async().unwrap() {
+                result::free_async(self.cu_device_ptr, self.device.stream).unwrap();
+            } else {
+                // result::free_sync(self.cu_device_ptr).unwrap();
```
coreylowman, line 200: Intended to be commented out?

Narsil: Sorry, I was trying things out because I still see some illegal address on compute_cap 5.2.
No it needs to be there !

src/driver/safe/alloc.rs
Narsil: I just wanted to keep it simple first, but quite easy to do.

Narsil:
@coreylowman The example is actually because this examples always fails to build. I started just kicking it out whenever I'm using tests.

I'll restore for this PR, but would be nice to have a real fix.
Same goes for the 2 multi device tests they are failing on single GPU machines which does not feel super great :)

src/driver/safe/alloc.rs
Narsil: Done.

src/driver/safe/core.rs
```diff
@@ -194,7 +204,11 @@ impl<T> Drop for CudaSlice<T> {
     fn drop(&mut self) {
         self.device.bind_to_thread().unwrap();
         unsafe {
-            result::free_async(self.cu_device_ptr, self.device.stream).unwrap();
+            if self.device.is_async {
+                result::free_async(self.cu_device_ptr, self.device.stream).unwrap();
+            } else {
+                // result::free_sync(self.cu_device_ptr).unwrap();
```
Narsil, line 210:
```suggestion
                result::free_sync(self.cu_device_ptr).unwrap();
```

coreylowman (approved):
lgtm

coreylowman:
@Narsil new version up on crates.io

PR #175, opened 2023-08-11
Fixing 07-build-workflow.

Narsil:
- `cargo test`  was broken for now because of this example (`build.rs` is not respected and `bindings.rs` is missing).
- This moves the example into it's own crate (https://github.com/rust-lang/cargo/issues/7816).
- Fixed the example by fixing `MyStruct` type and testing code.


Narsil:
@coreylowman

coreylowman:
Thank you!

PR #176, opened 2023-08-11
Making the tests pass on single GPU platforms.

Narsil:
- These tests fail on single GPU machine (which I guess is most devs setups).
- This adds a simple device count check before running the test.

Alternatives:

- Use a feature flag (seems overengineered).
- Add `#[ignore]` (but then we're discarding those tests basically).
- Use and environment variable.

I feel like the proposed approach is simple enough we actually get the testing run on valid platforms.
If there's a need for much more multi GPU tests, then maybe the solution should evolve.


Narsil:
@coreylowman

coreylowman:
Yep great minimal solution, thanks 👍

PR #177, opened 2023-08-11
Fix typo in docs about for_num_elems of LaunchConfig

mert-kurttutan:
Just noticed!


coreylowman:
Hah great catch, thanks!!

Issue #102, opened 2023-03-23
Add occupancy apis to driver/result and driver/safe


chelsea0x3b:
Detailed here https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__OCCUPANCY.html#group__CUDA__OCCUPANCY

Jark5455:
can we get an update on this? is there a pr?

coreylowman:
Addressed by pr above, thanks @Jark5455 !

PR #180, opened 2023-09-15
Minimum changes for occupancy API

Jark5455:
Hello,

Here is a draft pull request for #102. These are only the minimum changes (no docs, default parameters, etc..)

### DIFF
--- a/src/cublas/safe.rs
+++ b/src/cublas/safe.rs
@@ -61,61 +61,6 @@ impl Drop for CudaBlas {
     }
 }
 
-#[derive(Debug, Copy, Clone)]
-pub struct AxpyConfig<T> {
-    pub alpha: T,
-    pub n: c_int,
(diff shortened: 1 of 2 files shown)


coreylowman (changes requested):
Look good to me, can you run `cargo fmt` and then I can merge

Jark5455:
I believe it is failing because of a change that isn't made by me. The `Cudnn::new` function is returning an Arc to a Cudnn handle even though it shouldn't be able to since Cudnn handles are not thread safe.

coreylowman:
Actually you don't need to fix the Arc, I can fix that separately - Can you revert that change but keep the cargo fmt? I will manually merge even though the piplines fail 😄

Jark5455:
Sorry about that, im struggling with git a bit, I think its good now.

PR #182, opened 2023-10-02
feat: Add support for cublas Lt

OlivierDehaene:
Hello!
A fairly substantial pull request to add support for cublasLT. The key distinction between cublasLT and cublas lies in its epilogue system, which enables the fusion of kernels with bias and activation functions.


OlivierDehaene:
Hello!
FYI we use this in https://github.com/huggingface/text-embeddings-inference and it would be great to upstream it :)

coreylowman (approved):
Awesome addition, thanks for the PR!

PR #186, opened 2023-10-26
feat: add support for batch matrix-matrix product in cuBLASLt

OlivierDehaene:


coreylowman (approved):
Looks good, this is a breaking api change though so the next release will be a minor patch bump

PR #192, opened 2023-11-09
fix: fix cublaslt objects memory leak

OlivierDehaene:
Pretty big mistake on my side. The cuda objects were never destroyed.

### DIFF
--- a/src/cublaslt/safe.rs
+++ b/src/cublaslt/safe.rs
@@ -85,6 +85,69 @@ pub enum Activation {
     Gelu,
 }
 
+/// MatrixLayout helper type
+struct MatrixLayout {
+    handle: sys::cublasLtMatrixLayout_t,
+}
+
+impl MatrixLayout {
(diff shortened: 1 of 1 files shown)


src/cublaslt/safe.rs
```diff
@@ -85,6 +85,194 @@ pub enum Activation {
...
+        }
+        Ok(())
+    }
+}
+
+impl Drop for MatrixLayout {
+    fn drop(&mut self) {
+        // panic on failure
+        unsafe {
+            result::destroy_matrix_layout(self.handle).expect("Unable to destroy matrix layout")
```
OlivierDehaene, line 129: I'm potentially panicking on drop. Is that the correct way of dealing with failable functions in drop?

Narsil: This is how it's done elsewhere indeed.

Narsil:
@coreylowman FYI

Issue #196, opened 2023-12-15
Dynamic loading instead of linking


fayalalebrun:
Currently a binary depending on cudarc will fail to run when nvidia drivers are not installed. This means that programs which offer optional CUDA support need to distribute at least two binaries. The driver is also necessary for building the program, since otherwise the linking step will fail.

An alternative is to perform dynamic loading with a crate like `libloading`. `bindgen` already offers support for generating the required structures. In this way, the library will only be loaded when a function depending on CUDA is used, and the nvidia driver will no longer be required for building the binary.

Do you think this would be a desirable change? And do you see any point of still supporting strict dynamic linking if dynamic loading is supported?

coreylowman:
I'm going to work on adding this next - have become very interested in this feature. The main thing we'll need to figure out is if we still want to support dynamic & static linking in addition to dynamic loading.

coreylowman:
@Narsil  do you have any thoughts here? Curious how this could benefit candle

jedbrown:
It's quite common in the HPC world that interactive nodes do not have GPUs nor the drivers, but the compute nodes do. This is fine when just using `libcudart.so` because it loads the driver dynamically, and one can use `$CUDA_DIR/lib/stubs/` if your code calls the driver interface directly. I think similar/better usability is desirable for Rust, especially if it means we can distribute a static binary that can run on CPU machines without any CUDA libraries or devices, but can use devices when selected by run-time options on a machine that has them.

coreylowman:
Update on static linking: I think we don't necessarily need this feature anymore. I was able to run `cargo test` inside a docker image (one of the official nvidia ones) with some changes to LD_LIBRARY_PATH.

I think we can keep dynamic linking feature in the build.rs, but still use libloading in both loading/linking cases, and the OS should detect that we've already loaded the library in the dynamic linking case.

Narsil:
dynamic linking should still be the default imho as it's the most common way to distribute things (in a package manager you just need to depend on something). The kernel is able to do all the loading and finding things and also has some security model.
You rarely distribute a binary that supports multiple CPU arch. GPU is becoming more and more like a CPU imho in the sense that it's a core target for binaries in ML world.

Static linking is nice when you want to reduce the binary size and may be running on hosts that don't have cuda itself (quite niche but useful when you want to strip everything barebones, definitely removable).

dynamic loading is asking for trouble because of version issues in cuda, you might be loading a wrong version of the lib and making your program buggy without ever realizing.
[...]

PR #197, opened 2023-12-19
RFC: Add dynamic loading

fayalalebrun:
I explain the reasoning behind using dynamic loading instead of dynamic linking in #196. This is a quick implementation of dynamic loading only for the driver, created for the purpose of starting a discussion.

A few points I would like to ask about:
1. How to handle no-std. In fact, why does this crate support no-std environments? The nvidia driver seems to require a C standard library, so which system can run nvidia drivers but does not have access to the standard library?
2. How to handle static linking. In fact, why is static linking ever desirable with nvidia drivers? With nvidia drivers, userspace drivers must match kernel drivers, or they will not work. Therefore, the binary produced from static linking would only reliably work on the same machine which performed compilation. Since this machine already needed to have the dynamic libraries in order to statically link in the first place, it seems redundant to me.


coreylowman:
> How to handle no-std. In fact, why does this crate support no-std environments? The nvidia driver seems to require a C standard library, so which system can run nvidia drivers but does not have access to the standard library?

A big use case I had in mind for both cudarc/dfdx were for embedded devices where the std library is not necessarily needed. You raise a valid point about the nvidia driver requiring it. I think in this crate's case though, not linking against the std library could reduce binary sizes & memory sizes for us. So nvidia still might be using the C std library but cudarc wouldn't (well it would be implicitly through nvidia, but not directly).

As far as how to handle it if we move forward with dynamic loading, would have to reflect further on that.

I do like the idea of making optional cuda support binaries easier though - that is a use case that is hard to support in dfdx because everything is done with feature flags currently.

> In fact, why is static linking ever desirable with nvidia drivers?

Static linking was added in #104 #105, you can see the reasoning in the description for those PRs.

> With nvidia drivers, userspace drivers must match kernel drivers, or they will not work. Therefore, the binary produced from static linking would only reliably work on the same machine which performed compilation.

Yeah AFAIK most places that use cudarc do compile kernels on the machine they run on, so we haven't run into this issue (or at least no one has raised this issue here yet). The static kernels are very niche - it isn't a default feature flag, and I don't know how widely it is used.

coreylowman:
Implemented in #211

Issue #178, opened 2023-08-23
More cudnn ops


LaurentMazare:
Thanks for this amazing crate, it's been instrumental to candle. We've recently added a feature to use the cudnn conv2d which sped things up a lot compared to our handcrafted kernel, and would like to have cudnn support for more ops. Mostly interested in:
- Conv2d backprop.
- Conv1d forward + backward.
- Maybe flash-attention/softmax/...
Are there any plans to add these to the cudnn safe api? If not would you be ok with people making PR to add it?

coreylowman:
You can already do conv2d backprop with the existing methods (See https://github.com/coreylowman/dfdx/blob/main/src/tensor_ops/conv2d/cudnn_kernel.rs#L91).

Conv1d - I'm not sure this exists in cudnn? Not sure about flash attention existence either (or at least it didn't when I last checked cudnn)

But yes open to any contributions here!

LaurentMazare:
Ah great that the conv2d backward step is already there, we'll add it to candle.

For the conv1d, there is some support for Nd convolution I think, e.g. [in the cudarc ffi](https://github.com/coreylowman/cudarc/blob/4d956e2717b2c71714afb16598514f95a7ee994e/src/cudnn/sys.rs#L3230) so hopefully having a safe api around this would enable 1d convolutions.

For flash attention, I meant [this fused flash attn fprop](https://docs.nvidia.com/deeplearning/cudnn/developer-guide/index.html#flash-fused-multi-head-att-fprop), though I've actually never used it.

kckeiks:
Pooling would be great to have as well. It seems like it could have a similar api as conv.

kckeiks:
@coreylowman would you take a PR for adding nd pooling?

coreylowman:
@kckeiks of course, any and all prs welcome

PR #202, opened 2024-02-05
Safe ConvolutionND

bloodre:
## Summary
This PR introduces safe convolution operations for dimension N>=4, related safe APIs, tests, and deprecates Conv2d operations. Work was started after reading [this issue content](https://github.com/coreylowman/cudarc/issues/178).

### Content
- Safe wrapper for [cudnnSetFilterNdDescriptor](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnSetFilterNdDescriptor)
- Safe wrapper for [cudnnSetConvolutionNdDescriptor](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnSetConvolutionNdDescriptor)
- Create of alias for Conv2dDescriptor, Conv2dForward, Conv2dBackwardData, Conv2dBackwardFilter: ConvDescriptor, ConvForward, ConvBackwardData, ConvBackwardFilter. The original struct names have also been deprecated. The reason is beside "2d" being included in the struct name, their methods apply to convolution operation in any dimension (pick_algorithm, get_workspace_size, launch).
- Addition of tests: one for 3D convolution (using new APIs) and another for 1D conv (sample code with Conv2d operation)

### Regarding 1D convolution
As far as I understood from the CUDNN API reference, the ConvolutionNd are not used for 1D convolution:
- From [cudnnSetTensorDescriptor](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnSetTensorNdDescriptor): "Tensors are restricted to having at least 4 dimensions, and at most CUDNN_DIM_MAX dimensions (defined in cudnn.h). When working with lower dimensional data, it is recommended that the user create a 4D tensor, and set the size along unused dimensions to 1. ".
- [cudnnConvolutionForward](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnConvolutionForward) will return CUDNN_STATUS_BAD_PARAM if the input tensor descriptor and weight tensor description dimensions mismatch.  Using 3-dimensional filters is therefore not possible for convolutions.
As shown in the test "test_conv1d" that I added in src/cudnn/safe/mod.rs, a workaround is therefore to define a 4D tensor with the last dimension equal to 1, and to use 2D convolution operation with a flattened kernel.

### Note
- It might be useful to create some utility function to generate the stride arrays of Nd tensors from the dimensions.

### DIFF
--- a/src/cudnn/safe/conv.rs
+++ b/src/cudnn/safe/conv.rs
@@ -92,7 +92,7 @@ pub struct ConvDescriptor<T> {
     pub(crate) marker: PhantomData<T>,
 }
 #[deprecated(note = "use ConvDescriptor instead. This will be removed in future versions")]
-type Conv2dDescriptor<T> = ConvDescriptor<T>;
+pub type Conv2dDescriptor<T> = ConvDescriptor<T>;
 
 impl Cudnn {
     /// Creates a conv2d descriptor.
@@ -205,7 +205,7 @@ pub struct ConvForward<'a, X: CudnnDataType, C: CudnnDataType, Y: CudnnDataType>
     pub y: &'a TensorDescriptor<Y>,
 }
 #[deprecated(note = "use ConvForward instead. This will be removed in future versions")]
-type Conv2dForward<T> = ConvForward<T>;
+pub type Conv2dForward<X, C, Y> = ConvForward<X, C, Y>;
 
 impl<'a, X: CudnnDataType, C: CudnnDataType, Y: CudnnDataType> ConvForward<'a, X, C, Y> {
     /// Picks the fastest algorithm from all available cuDNN algorithms based on cudnn heuristics.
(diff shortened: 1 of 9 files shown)


src/cudnn/safe/conv.rs
```diff
@@ -31,6 +31,48 @@ impl Cudnn {
...
+                format,
+                dims.len() as std::ffi::c_int,
+                dims.as_ptr(),
+            )
+        }?;
+        Ok(desc)
+    }
+
+    /// Create a filter 1d descriptor.
+    pub fn create_1d_filter<T: CudnnDataType>(
```
coreylowman, line 60:
```suggestion
    /// Create a filter 3d descriptor.
    pub fn create_3d_filter<T: CudnnDataType>(
```

coreylowman:
Looks good, thanks for this!

PR #204, opened 2024-02-07
Implemented LaunchAsync for raw ptr arrays

jafioti:
I'm working with runtime-generated kernels that can take in varying amounts of inputs. The number of inputs isn't known at compile time, so I build up a vec of c_void pointers to the inputs. This didn't work with launch, so I implemented LaunchAsync trait for arrays of pointers.


jafioti:
@coreylowman Would you be able to take a look at this?

coreylowman (approved):
Looks good, thanks for tagging

PR #205, opened 2024-02-16
Fixed Test Example 07

Fiend-Star:
Test Example 07 was failing due to version mismatch of cudarc, so i made it dynamic  and the assert was failing because we were incrementing the values by 1 therefore the final values should be checked if its equal to 2 rather than 1


coreylowman:
Thanks!

PR #208, opened 2024-02-29
Fix `test_launch_with_64bit`. Update deps

DanikVitek:
- Replace kernel literals with raw strings
- Fix `test_launch_with_64bit`
- Update dependencies
- Bump version to 0.10.1


DanikVitek:
I'd also suggest to include the `stdint.h` header for C99 fixed-size integers. Though, I'm not sure exactly how.

PR #210, opened 2024-03-13
Adding cuda toolkit versions 12.0/12.1/12.2 of sys & feature flags

### LINKED ISSUE #201: Best way to support different CUDA version?
If I want to use a different CUDA version than the bindgen is currently built against, should I just rerun bindgen on my system?

Would there be interest in a PR to make this crate dynamically generate bindings based on the cuda installation on the developer's system?

coreylowman:
Closes #201


Cargo.toml
```diff
@@ -24,7 +24,9 @@ keywords = [
 features = ["ci-check", "f16", "cudnn"]
 
 [features]
-default = ["std", "driver", "nvrtc", "cublas", "curand"]
+default = ["std", "driver", "nvrtc", "cublas", "curand", "cuda_version_11_8"]
+cuda_version_11_8 = []
+cuda_version_12_2 = []
```
coreylowman, line 29: These could probably be set automatically in build.rs based on whatever CUDA is on the machine

PR #213, opened 2024-03-25
Add CUDA 12.4 (#212)

coreylowman:
TODO:
- [ ] Generate bindings for cudnn
- [ ] Genearte bindings for nccl


bitemyapp:
My PR over at #227 should include cudnn and nccl unless I've missed something.

coreylowman:
Fixed in a separate PR

PR #214, opened 2024-03-26
support cuda toolkit version: 11.7

wenhaozhao:
This PR adds support for CUDA Toolkit 11.7


corey-lambda:
This looks great! Just will need to add nccl & cudnn.

You can extract these inside `docker pull nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04` if you don't have access to nccl/cudnn on whatever machine you produced these on!

coreylowman:
This looks great! Just will need to add nccl & cudnn.

You can extract these inside `docker pull nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04` if you don't have access to nccl/cudnn on whatever machine you produced these on!

wenhaozhao:
> This looks great! Just will need to add nccl & cudnn.
> 
> You can extract these inside `docker pull nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04` if you don't have access to nccl/cudnn on whatever machine you produced these on!

ok, thanks!
I've added two files about **nccl** & **cudnn**
- 1 src/cudnn/sys/sys_11070.rs
- 2 src/nccl/sys/sys_11070.rs

coreylowman:
Yay perfect, thanks

coreylowman:
Hmm looks like there were some pretty breaking changes for cudnn/nccl on 11.7 (getting errors when trying to compile for that version in the safe apis). Will need to think about that

PR #229, opened 2024-05-03
Moving cublasHgemm link to dynamic loading

### LINKED ISSUE #228: Error searching for cublasHgemmStridedBatched
I get an error when compiling. I am on CUDA 12.2:
```
undefined reference to `cublasHgemmStridedBatched'
          collect2: error: ld returned 1 exit status
          
  = note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)
```

coreylowman:
Fixes #228


coreylowman:
@jafioti I wasn't able to reproduce your error, but can you check out this branch and test with it?

I couldn't tell if you were actually trying to use those functions, but it's recommended to use the `result::gemm_ex` and `result::gemm_strided_batched_ex` for f16 gemm. Perhaps we should just remove these hgemm specific functions since they aren't included in the bindgen export. They aren't used anywhere in the safe api currently though.

jafioti:
Sorry for the late response, this indeed does fix the issue. Thanks!

Issue #246, opened 2024-06-06
Unable to find nvrtc lib under the names ["nvrtc", "nvrtc64", "nvrtc64_12", "nvrtc64_125", "nvrtc64_125_0", "nvrtc64_120_5", "nvrtc64_10"]. Please open GitHub issue.


brandonros:
```
PS C:\Users\Brandon\Desktop\cuda_fun> cargo run
   Compiling cuda_fun v0.1.0 (C:\Users\Brandon\Desktop\cuda_fun)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.96s
     Running `C:\Users\Brandon\.cargo\.target\debug\cuda_fun.exe`
thread 'main' panicked at C:\Users\Brandon\.cargo\registry\src\index.crates.io-6f17d22bba15001f\cudarc-0.11.4\src\nvrtc\sys\mod.rs:51:9:
Unable to find nvrtc lib under the names ["nvrtc", "nvrtc64", "nvrtc64_12", "nvrtc64_125", "nvrtc64_125_0", "nvrtc64_120_5", "nvrtc64_10"]. Please open GitHub issue.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `C:\Users\Brandon\.cargo\.target\debug\cuda_fun.exe` (exit code: 101)
```

![image](https://github.com/coreylowman/cudarc/assets/8949910/6ba78d59-f907-4b46-b802-c054be7b89f8)

No clue why CUDA v12.5 would have `nvrtc64_120_0.dll` DLL but guess we need to support that?

Guessing it needs to be added here: https://github.com/coreylowman/cudarc/blob/6fa8132e87c66efd56c9580235af66559a513d2e/src/lib.rs#L97

brandonros:
closed via https://github.com/coreylowman/cudarc/pull/250#event-13121922611

Issue #219, opened 2024-04-16
Wrong library filename used for CUDA driver on Windows


jhuang97:
I was trying to run the examples, and `let dev = CudaDevice::new(0)?;` was causing a LoadLibraryExW error for me.

To fix this for myself, I replaced 
`LIB.get_or_init(|| Lib::new(libloading::library_filename("cuda")).unwrap())`
in `src/driver/sys/mod.rs` with 
`LIB.get_or_init(|| Lib::new(libloading::library_filename("nvcuda")).unwrap())`.

I think this is because I am on Windows and the CUDA driver library is called `nvcuda.dll`.

Edit: I also needed to have
`LIB.get_or_init(|| Lib::new(libloading::library_filename("nvrtc64_120_0")).unwrap())`
in `src/nvrtc/sys/mod.rs`.

coreylowman:
@jhuang97 can you also try this on cudarc version 0.10.0? I'm wondering if this error was introduced with the swap to dynamic loading. 0.10.0 does static linking only, so if this was introduced recently you should be able to compile on that version. If you aren't able to compile on 0.10.0, then that might be an edge case system setup and we can work to get a workaround.

jhuang97:
I tried running the examples on cudarc version 0.10.0.  Everything is fine on that version.

coreylowman:
@jhuang97 thanks! If you enable the dynamic-linking feature flag with 0.11.0 (or current main branch), do things work for you? So `cargo run -F dynamic-linking <whatever command you were runnign to test things>`.

I think we can definitely add a fallback for loading nvcuda if loading cuda doesn't work:
```rust
LIB.get_or_init(|| {
        Lib::new(libloading::library_filename("cuda"))
            // checking for nvcuda in case of error on windows. See issue #219.
            .or_else(|e| Lib::new(libloading::library_filename("nvcuda")))
            .unwrap()
    })
```
I'm a little hesitant about nvrtc one because it seems like they are encoding some versioning information into the dll name which could get really complicated.

PR #247, opened 2024-06-06
add NVRTC_LIB_OVERRIDE

brandonros:
closes https://github.com/coreylowman/cudarc/issues/246

slightly related to https://github.com/coreylowman/cudarc/issues/219

### DIFF
--- a/src/nvrtc/sys/mod.rs
+++ b/src/nvrtc/sys/mod.rs
@@ -43,7 +43,7 @@ pub unsafe fn lib() -> &'static Lib {
     LIB.get_or_init(|| {
         match std::env::var("NVRTC_LIB_OVERRIDE") {
             Ok(nvrtc_lib_override) => {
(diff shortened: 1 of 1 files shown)


brandonros:
image

works for me

coreylowman:
I think a more straightforward change would be to add `std::format!("{lib_name}{pointer_width}_{major}0_0"),` to the list of choices under https://github.com/coreylowman/cudarc/blob/6fa8132e87c66efd56c9580235af66559a513d2e/src/lib.rs#L115C9-L115C68.

Just so its consistent with how it works now. 

No idea why windows is so different for cuda world lol.

coreylowman (changes requested):
As I mentioned in the other comment, let's just add another option to the list in `get_lib_name_candidates`, so all the libraries get access to this improvement. No idea what window libs have this issue, so better fix it for all of them pre-emptively.

brandonros:
moving to https://github.com/coreylowman/cudarc/pull/247

PR #251, opened 2024-06-11
[Breaking][Waiting to Merge] add support for passing options to nvrtc

brandonros:


coreylowman:
Thanks for the PR.

Do you know if there are other compile options we should add? This is a breaking change (we are adding a pub field to a pub struct) so might be a good time to add other options as well, or bundle with other breaking changes

brandonros:
went for a more generic approach. thanks for taking a look, no rush on merging/happy to throw away if needed, glad to see package is maintained. super fun to use, appreciate all of your work on this

PR #254, opened 2024-06-12
Implement Send + Sync for CudaStream

sebhtml:
Hi !

First, thank you for the cudarc project. It's a nice piece of engineering.

In my [Novigrad](https://github.com/sebhtml/novigrad/tree/main/src/neural_machine/scheduler) project, I use cudarc.

My models are declared like in PyTorch (but in Rust).
Then, I generate [instructions with opcodes and operands](http://dskernel.blogspot.com/2024/05/adam-optimizer-perplexity-and-bunch-of.html).

Using a dependency analysis, I assign each instruction to a logical stream. Logical streams have dependencies too.
At runtime, a logical stream is mapped to a physical CUDA stream.

For example, for multi-head attention, with 12 attention heads, the dependency analysis figures out that each head can be executed on a separate logical stream. Then the logical stream for Concat depends on the 12 logical streams of the heads.

I use a scheduler with a controller and execution units in a multi-threaded implementation.
It requires CudaStream to implement Send + Sync.

I ran the tests (cargo test --release). I don't currently have cuddn and nccl, so those tests failed.

```
failures:

---- cudnn::safe::tests::test_create_descriptors stdout ----
thread 'cudnn::safe::tests::test_create_descriptors' panicked at src/cudnn/sys/mod.rs:61:9:
Unable to find cudnn lib under the names ["cudnn", "cudnn64", "cudnn64_11", "cudnn64_115", "cudnn64_115_0", "cudnn64_110_5", "cudnn64_10", "cudnn64_110_0"]. Please open GitHub issue.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- cudnn::safe::tests::test_conv2d_pick_algorithms stdout ----
thread 'cudnn::safe::tests::test_conv2d_pick_algorithms' panicked at src/cudnn/sys/mod.rs:61:9:
Unable to find cudnn lib under the names ["cudnn", "cudnn64", "cudnn64_11", "cudnn64_115", "cudnn64_115_0", "cudnn64_110_5", "cudnn64_10", "cudnn64_110_0"]. Please open GitHub issue.

---- cudnn::safe::tests::test_conv1d stdout ----
thread 'cudnn::safe::tests::test_conv1d' panicked at src/cudnn/sys/mod.rs:61:9:
[...]

PR #261, opened 2024-06-20
Add documentation on how to synchronize non-default streams

sebhtml:
**Problem statement**

As a Developer, when I read https://docs.rs/cudarc/latest/cudarc/driver/safe/struct.CudaStream.html, it is not clear how I can use cudarc to synchronize work across non-default CUDA streams.

**Solution**

In this PR, I added documentation for how to synchronize many non-default streams.

**Context**

In my previous PR https://github.com/coreylowman/cudarc/pull/254, I thought that CudaStream streams were thread-safe. 
My understanding was incorrect.

@coreylowman then kindly pointed that usually one achieves on-device concurrency by launching from the same host thread several CUDA kernels on several non-default non-blocking CUDA streams. So I read the CUDA documentation about streams to determine how to achieve this with cudarc.

cudarc uses the [CUDA events API](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html) to synchronize work.

**Use case**

I use cudarc with this approach in my novigrad project ([GPU Scheduler](https://github.com/sebhtml/novigrad/blob/main/src/neural_machine/schedulers/gpu_scheduler/mod.rs), [DeviceStream](https://github.com/sebhtml/novigrad/blob/main/src/devices/stream.rs)) and it works very well apparently.

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -458,7 +458,7 @@ unsafe impl Send for CudaFunction {}
 unsafe impl Sync for CudaFunction {}
 
 /// A wrapper around [sys::CUstream] that safely ensures null stream is synchronized
-/// upon the completion of this streams work.
+/// upon the completion of this stream's work.
 ///
 /// Create with [CudaDevice::fork_default_stream].
 ///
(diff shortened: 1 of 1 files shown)


coreylowman:
Thanks for this!

PR #283, opened 2024-08-07
Allow CUDA 12.6

paulstansifer:
I still haven't gotten `cudarc` properly set up on my machine, so this change is totally untested.


paulstansifer:
Never mind! The compile errors I was getting were because there's more to supporting a new version than I realized at first!

PR #303, opened 2024-12-21
Elazarl/feature/cuuuid

elazarl:
Expose cuDeviceGetName from device


elazarl:
@coreylowman 

Would `uuid` as the function name be acceptable?

Why do we have `get_name` when rust getters seems not to have the `get_` prefix?

elazarl:
Anyone available to review? @coreylowman

coreylowman (approved):
lgtm! nice addition

Issue #307, opened 2025-01-07
error[E0425]: cannot find function `cudaImportExternalMemory` in module `sys`


EricLBuehler:
Hi @coreylowman!

We are running into a strange error when attempting to compile cudarc 0.12.2 on Windows.

This is the error:
```
error[E0422]: cannot find struct, variant or union type `cudaExternalMemoryHandleDesc_st__bindgen_ty_1` in module `sys`
    --> C:\Users\ericl\.cargo\registry\src\index.crates.io-6f17d22bba15001f\cudarc-0.12.2\src\runtime\result.rs:988:26
     |
988  |             handle: sys::cudaExternalMemoryHandleDesc_st__bindgen_ty_1 {
     |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a union with a similar name exists: `cudaExternalMemoryHandleDesc__bindgen_ty_1`
     |
    ::: C:\Users\ericl\.cargo\registry\src\index.crates.io-6f17d22bba15001f\cudarc-0.12.2\src\runtime\sys\sys_12060.rs:4388:1
     |
4388 | pub union cudaExternalMemoryHandleDesc__bindgen_ty_1 {
     | ---------------------------------------------------- similarly named union `cudaExternalMemoryHandleDesc__bindgen_ty_1` defined here    

error[E0422]: cannot find struct, variant or union type `cudaExternalMemoryHandleDesc_st__bindgen_ty_1__bindgen_ty_1` in module `sys`
    --> C:\Users\ericl\.cargo\registry\src\index.crates.io-6f17d22bba15001f\cudarc-0.12.2\src\runtime\result.rs:989:29
     |
989  |                 win32: sys::cudaExternalMemoryHandleDesc_st__bindgen_ty_1__bindgen_ty_1 {
     |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a struct with a similar name exists: `cudaExternalMemoryHandleDesc__bindgen_ty_1__bindgen_ty_1`
     |
    ::: C:\Users\ericl\.cargo\registry\src\index.crates.io-6f17d22bba15001f\cudarc-0.12.2\src\runtime\sys\sys_12060.rs:4395:1
     |
4395 | pub struct cudaExternalMemoryHandleDesc__bindgen_ty_1__bindgen_ty_1 {
[...]

PR #308, opened 2025-01-07
Fix build on Windows

EricLBuehler:
Refs EricLBuehler/mistral.rs#999.

Fixes Windows build which currently crashes on master with:
```
cargo build --features std,cublas,cublaslt,curand,driver,nvrtc,f16,cuda-version-from-build-system,dynamic-linking
```

@coreylowman If you could review this that would be great!


EricLBuehler:
Closing as this is actually a duplicate fix for #307!

PR #335, opened 2025-02-23
[WIP] Refactoring CudaStream & add CudaStream::launch/launch_cooperative

coreylowman:
# TODOS

- [ ] Decide on what methods are deprecated and what should be removed. Should this PR be breaking?
- [ ] Update documentation
- [ ] Update examples
- [ ] Updating tests
- [ ] How to soundly manage CudaSlices moving between streams
    - Previously since it was not possible to change the stream of a CudaDevice, all CudaSlices created would be on the same stream. Now it is possible to create a CudaSlice on one stream, and then use another stream to read/modify it. This can be very unsound if you don't do stream synchronization between the streams it belongs too.
    - The first iteration of this update changes `CudaSlice` to track its creation stream
    - It is easy to check if the reading/modifying stream is the same stream that the `CudaSlice` was created on.
    - It is NOT clear how to update which stream a `CudaSlice` belongs to
    - It is NOT clear if cudarc should even try to automatically manage this, or just expose things to make it easier to do for the developer.
    - A motivating example would be: allocate a CudaSlice on a stream, Split the CudaSlice into 2 CudaViewMuts, do a memset_zeros on 2 additional streams.
    - Another example is that the `impl Drop for CudaSlice` currently schedules the free_async on the stream that the CudaSlice was created on

# Background

Currently there's no way to have an existing CudaDevice use a separate CudaStream for things like allocation/memcpy/memset operations. There's CudaDevice::new_with_stream, but that won't share any loaded kernels that had been previously loaded.

Additionally, the current LaunchAsync trait is a bit too complicated and only supports a small number of compile time arguments.

# API Changes

## Allocation & copy apis

- Add `CudaDevice::default_stream() -> CudaStream`
- Add `CudaStream::null()`, `CudaStream::alloc()`, `CudaStream::alloc_zeros()`, `CudaStream::memset_zeros()`, `CudaStream::dtod_copy()`, `CudaStream::htod_copy()`, `CudaStream::htod_copy_into()`, `CudaStream::htod_sync_copy()`, `CudaStream::htod_sync_copy_into()`, `CudaStream::dtoh_sync_copy()`, `CudaStream::dtoh_sync_copy_into()`, `CudaStream::sync_reclaim()`, `CudaStream::synchronize()`

The intention after these changes is that the main way you submit jobs to cuda devices is through a `CudaStream` instead of a `CudaDevice`. Since `CudaStream` contains an `Arc<CudaDevice>`, in most cases you can just replace `CudaDevice` with `CudaStream` in downstream library code.

After this PR, **all `CudaDevice` calls just call `self.default_stream().<method>()`**

## Kernel launching API

- Add `CudaStream::launch()` and `CudaStream::launch_cooperative()`.

These methods have a much simpler API for passing kernel params:

```rust
pub unsafe fn launch<const NUM_ARGS: usize>(
    &self,
    func: CudaFunction,
    cfg: LaunchConfig,
    args: [&dyn DeviceRepr; NUM_ARGS],
) -> Result<(), result::DriverError>;
```

Here we've replaced tuples with a compile time sized array with trait objects that implement `DeviceRepr`. This is essentially what LaunchAsync did under the hood, with the following benefits:
1. It supports ANY number of arguments, without complicated trait implementations (we recently bumped the number of tuple args to 32)
2. It becomes much clearer how to launch things on separate streams (just use a separate stream object!)

## Stream creation apis

- `CudaDevice::new_stream()` creates a new stream without any synchronization with the device's current stream.
- `CudaStream::fork()` replaces the existing `CudaDevice::fork_default_stream()` and makes it more generally applicable. When you fork a stream, it will create an event at creation & at drop to synchronize with the stream it was forked from.

## Stream synchronization apis

- `CudaStream::wait_for()` enables synchronization between two streams. Replaces `CudaDevice::wait_for()` and `CudaStream::wait_for_default()`

### DIFF
--- a/src/cudnn/safe/core.rs
+++ b/src/cudnn/safe/core.rs
@@ -20,7 +20,6 @@ impl Cudnn {
     pub fn new(device: Arc<CudaDevice>) -> Result<Arc<Self>, CudnnError> {
         device.bind_to_thread().unwrap();
         let handle = result::create_handle()?;
-        unsafe { result::set_stream(handle, device.stream as *mut _) }?;
         Ok(Arc::new(Self { handle, device }))
     }
 
@@ -32,8 +31,11 @@ impl Cudnn {
     /// write to the same memory address.
     pub unsafe fn set_stream(&self, opt_stream: Option<&CudaStream>) -> Result<(), CudnnError> {
         match opt_stream {
-            Some(s) => result::set_stream(self.handle, s.stream as *mut _),
-            None => result::set_stream(self.handle, self.device.stream as *mut _),
+            Some(s) => {
+                assert_eq!(self.device.ordinal(), s.device.ordinal());
+                result::set_stream(self.handle, s.cu_stream as *mut _)
+            }
+            None => result::set_stream(self.handle, std::ptr::null_mut()),
         }
     }
 }
--- a/src/driver/safe/alloc.rs
+++ b/src/driver/safe/alloc.rs
@@ -1,6 +1,6 @@
 use crate::driver::{result, sys};
 
-use super::core::{CudaDevice, CudaSlice, CudaView, CudaViewMut};
+use super::core::{CudaDevice, CudaSlice, CudaStream, CudaView, CudaViewMut};
 use super::device_ptr::{DevicePtr, DevicePtrMut, DeviceSlice};
 
 use std::{marker::Unpin, pin::Pin, sync::Arc, vec::Vec};
@@ -17,6 +17,10 @@ pub unsafe trait DeviceRepr {
     #[inline(always)]
     fn as_kernel_param(&self) -> *mut std::ffi::c_void {
         self as *const Self as *mut _
+    }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        None
     }
 }
 
@@ -40,10 +44,25 @@ unsafe impl DeviceRepr for half::f16 {}
 #[cfg(feature = "f16")]
 unsafe impl DeviceRepr for half::bf16 {}
 
+unsafe impl<T: DeviceRepr> DeviceRepr for CudaSlice<T> {
+    #[inline(always)]
+    fn as_kernel_param(&self) -> *mut std::ffi::c_void {
+        (&self.cu_device_ptr) as *const sys::CUdeviceptr as *mut std::ffi::c_void
+    }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        Some(&self.stream)
+    }
+}
+
 unsafe impl<T: DeviceRepr> DeviceRepr for &mut CudaSlice<T> {
     #[inline(always)]
     fn as_kernel_param(&self) -> *mut std::ffi::c_void {
         (&self.cu_device_ptr) as *const sys::CUdeviceptr as *mut std::ffi::c_void
+    }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        Some(&self.stream)
     }
 }
 
@@ -52,6 +71,21 @@ unsafe impl<T: DeviceRepr> DeviceRepr for &CudaSlice<T> {
     fn as_kernel_param(&self) -> *mut std::ffi::c_void {
         (&self.cu_device_ptr) as *const sys::CUdeviceptr as *mut std::ffi::c_void
     }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        Some(&self.stream)
+    }
+}
+
+unsafe impl<T: DeviceRepr> DeviceRepr for CudaView<'_, T> {
+    #[inline(always)]
+    fn as_kernel_param(&self) -> *mut std::ffi::c_void {
+        (&self.ptr) as *const sys::CUdeviceptr as *mut std::ffi::c_void
+    }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        Some(&self.stream)
+    }
 }
 
 unsafe impl<T: DeviceRepr> DeviceRepr for &CudaView<'_, T> {
@@ -59,12 +93,31 @@ unsafe impl<T: DeviceRepr> DeviceRepr for &CudaView<'_, T> {
     fn as_kernel_param(&self) -> *mut std::ffi::c_void {
         (&self.ptr) as *const sys::CUdeviceptr as *mut std::ffi::c_void
     }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        Some(&self.stream)
+    }
+}
+
+unsafe impl<T: DeviceRepr> DeviceRepr for CudaViewMut<'_, T> {
+    #[inline(always)]
+    fn as_kernel_param(&self) -> *mut std::ffi::c_void {
+        (&self.ptr) as *const sys::CUdeviceptr as *mut std::ffi::c_void
+    }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        Some(&self.stream)
+    }
 }
 
 unsafe impl<T: DeviceRepr> DeviceRepr for &mut CudaViewMut<'_, T> {
     #[inline(always)]
     fn as_kernel_param(&self) -> *mut std::ffi::c_void {
         (&self.ptr) as *const sys::CUdeviceptr as *mut std::ffi::c_void
+    }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        Some(&self.stream)
     }
 }
 
@@ -92,6 +145,7 @@ impl CudaDevice {
     /// - `cu_device_ptr` must space for `len * std::mem::size_of<T>()` bytes
     /// - The memory may not be valid for type `T`, so some sort of memset operation
     ///   should be called on the memory.
+    #[deprecated]
     pub unsafe fn upgrade_device_ptr<T>(
         self: &Arc<Self>,
         cu_device_ptr: sys::CUdeviceptr,
@@ -100,7 +154,30 @@ impl CudaDevice {
         CudaSlice {
             cu_device_ptr,
             len,
-            device: self.clone(),
+            stream: self.default_stream(),
+            host_buf: None,
+        }
+    }
+}
+
+impl CudaStream {
+    /// Creates a [CudaSlice] from a [sys::CUdeviceptr]. Useful in conjunction with
+    /// [`CudaSlice::leak()`].
+    ///
+    /// # Safety
+    /// - `cu_device_ptr` must be a valid allocation
+    /// - `cu_device_ptr` must space for `len * std::mem::size_of<T>()` bytes
+    /// - The memory may not be valid for type `T`, so some sort of memset operation
+    ///   should be called on the memory.
+    pub unsafe fn upgrade_device_ptr<T>(
+        &self,
+        cu_device_ptr: sys::CUdeviceptr,
+        len: usize,
+    ) -> CudaSlice<T> {
+        CudaSlice {
+            cu_device_ptr,
+            len,
+            stream: self.clone(),
             host_buf: None,
         }
     }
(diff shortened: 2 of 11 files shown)


src/cudnn/safe/core.rs
```diff
@@ -32,8 +31,11 @@ impl Cudnn {
     /// write to the same memory address.
     pub unsafe fn set_stream(&self, opt_stream: Option<&CudaStream>) -> Result<(), CudnnError> {
         match opt_stream {
-            Some(s) => result::set_stream(self.handle, s.stream as *mut _),
-            None => result::set_stream(self.handle, self.device.stream as *mut _),
+            Some(s) => {
+                assert_eq!(self.device.ordinal(), s.device.ordinal());
```
coreylowman, line 35: TODO should this assert or return an error? is setting the stream to a different device valid? Same comment for other set_stream methods

src/driver/safe/core.rs
```diff
@@ -559,68 +505,96 @@ unsafe impl Sync for CudaFunction {}
 /// See [6.6. Event Management](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html)
 /// See [Out-of-order execution](https://en.wikipedia.org/wiki/Out-of-order_execution)
 /// See [Dependence analysis](https://en.wikipedia.org/wiki/Dependence_analysis)
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct CudaStream {
-    pub stream: sys::CUstream,
```
coreylowman, line 564: TODO renaming this field would make this a breaking PR

src/driver/safe/core.rs
```diff
@@ -559,68 +505,96 @@ unsafe impl Sync for CudaFunction {}
...
             result::stream::wait_event(
-                self.stream,
-                self.device.event,
+                self.cu_stream,
+                event,
                 sys::CUevent_wait_flags::CU_EVENT_WAIT_DEFAULT,
-            )
+            )?;
+            // TODO test that this is safe
+            result::event::destroy(event)
```
coreylowman, line 575: The cuda docs say
> An event may be destroyed before it is complete (i.e., while [cuEventQuery()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EVENT.html#group__CUDA__EVENT_1g6f0704d755066b0ee705749ae911deef) would return [CUDA_ERROR_NOT_READY](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1ggc6c391505e117393cc2558fff6bfc2e9edd9cef666ce620352e619a36b6c3f34)). In this case, the call does not block on completion of the event, and any associated resources will automatically be released asynchronously at completion.

My interpretation of this is that we can destroy an event and it will still do the expected synchronization. Just need to test this

src/driver/safe/device_ptr.rs
```diff
@@ -1,8 +1,9 @@
 use crate::driver::sys;
 
-use super::core::{CudaSlice, CudaView, CudaViewMut};
+use super::core::{CudaSlice, CudaStream, CudaView, CudaViewMut};
 
 pub trait DeviceSlice<T> {
+    fn stream(&self) -> &CudaStream;
```
coreylowman, line 6: This is a breaking change. We really only need this if we are tracking what stream a CudaSlice/CudaView/CudaViewMut is owned by

src/driver/safe/alloc.rs
```diff
@@ -18,6 +18,10 @@ pub unsafe trait DeviceRepr {
     fn as_kernel_param(&self) -> *mut std::ffi::c_void {
         self as *const Self as *mut _
     }
+    #[inline(always)]
+    fn maybe_stream(&self) -> Option<&'_ CudaStream> {
+        None
```
coreylowman, line 23: Not a breaking change since there's a default impl, but not quite sure if we need this. This would be used in CudaStream::launch/CudaStream::launch_cooperative to check if the args belong to the launching stream

src/driver/safe/alloc.rs
```diff
@@ -168,16 +434,15 @@ impl CudaDevice {
...
-        if self.is_async {
-            unsafe {
-                result::memset_d8_async(*dst.device_ptr_mut(), 0, dst.num_bytes(), self.stream)
-            }
-        } else {
-            unsafe { result::memset_d8_sync(*dst.device_ptr_mut(), 0, dst.num_bytes()) }
+        self.device.bind_to_thread()?;
+        if dst.stream().cu_stream != self.cu_stream {
+            todo!();
+        }
```
coreylowman, line 443: This is an example of where we need to decide what to do if the launching stream is different from the CudaSlice's stream. Why this is dangerous:

1. Allocate a CudaSlice on stream A
2. Create stream B NOT synchronized with stream A
3. Call memset_zeros on stream B
4. The CudaSlice is dropped, and a free_async is scheduled on stream A (as of creating this PR)

If the CudaSlice is dropped before memset_zero is executed, we'd have a use after free error.

There are a couple of options we could do here:
1. Nothing - the user needs to track this
2. Warn somehow (we don't have a way to warn atm in cudarc)
3. Return an Err
4. panic
5. Change `dst.stream()` to the stream we are calling memset_zeros with (though we'd need to think if this works if we are using multiple streams).
6. Do some sort of stream synchronization to ensure the CudaSlice doesn't get dropped before memset_zero happens

serhii-havrylov: Hi, I just wanted to chime in regarding this. Currently, I am using your crate with some additional extensions that I've built on top (proper event management, graph capture, stream interactions, etc.), and I wanted to mention that it seems unlikely you would be able to provide a zero-cost abstraction for making all CUDA-related interactions safe.

At present, I track everything related to device synchronization myself using unsafe functions by explicitly specifying which invariants a given function expects to be upheld by the caller. However, it is possible to improve the experience and prevent trivial errors by making a few changes:

Whenever the GPU driver needs to modify resource handlers, Rust should take mutable references to the corresponding Rust abstractions. For example, in my code, I have the following async launcher for a CUDA function:
```rust
impl LaunchAsync<&mut [*mut c_void]> for CudaFunction {
    unsafe fn launch_on_stream(
        &self,
        stream: &mut CudaStream,
        cfg: LaunchConfig,
        params: &mut [*mut c_void],
    ) -> Result<(), result::DriverError> {
        self.device().bind_to_thread()?;
        result::launch_kernel(
            self.cu_function(),
            cfg.grid_dim,
            cfg.block_dim,
            cfg.shared_mem_bytes,
            stream.stream,
            params,
        )
    }
}
```
as you can see `stream` is `&mut CudaStream`. However, in your library, it is currently implemented as follows:
```rust
    unsafe fn launch_on_stream(
        self,
        stream: &CudaStream,
        cfg: LaunchConfig,
        args: &mut [*mut std::ffi::c_void],
    ) -> Result<(), result::DriverError> {
        self.par_launch_async_impl(stream, cfg, args)
    }
```
This allows a reference to stream to be easily given away to other parts of the code, preventing exclusive access to this resource. As a result, work could be erroneously submitted to the stream without realizing it. This is a perfect use case for the Rust borrow checker—it ensures that only one entity is responsible for computation orchestration.

The same principle applies to other resources. For example:
```rust
use cudarc::cublas::result::CublasError;
use cudarc::cublas::CudaBlas;
use cudarc::driver::CudaStream;

pub trait SetStream {
    fn set_stream(&mut self, stream: &mut CudaStream) -> Result<(), CublasError>;
}

impl SetStream for CudaBlas {
    fn set_stream(&mut self, stream: &mut CudaStream) -> Result<(), CublasError> {
        unsafe { (self as &CudaBlas).set_stream(Some(stream)) }
    }
}
```
Here, you are explicit about the fact that you are modifying the CudaBlas handler and mutably borrowing CudaStream because all future work initiated through the cuBLAS handler will be submitted to the stream. This prevents users from accidentally launching work on the stream without proper synchronization.
another example:
```rust
use cudarc::driver::{result, sys, CudaDevice, DriverError};
use std::sync::Arc;

pub struct CudaEvent {
    event: sys::CUevent,
    pub(crate) device: Arc<CudaDevice>,
}

impl CudaEvent {
    pub fn new(device: Arc<CudaDevice>) -> Result<Self, DriverError> {
        device.bind_to_thread()?;
        let event = result::event::create(sys::CUevent_flags::CU_EVENT_DISABLE_TIMING)?;
        Ok(Self { event, device })
    }

    pub fn event(&self) -> sys::CUevent {
        self.event
    }
}

impl Drop for CudaEvent {
    fn drop(&mut self) {
        unsafe { result::event::destroy(self.event) }.unwrap();
    }
}

pub trait Record {
    unsafe fn record(&mut self, cuda_stream: &CudaStream) -> Result<(), DriverError>;
}

impl Record for CudaEvent {
    unsafe fn record(&mut self, cuda_stream: &CudaStream) -> Result<(), DriverError> {
        self.device().bind_to_thread()?;
        result::event::record(self.event(), cuda_stream.stream)
    }
}

pub trait WaitFor {
    unsafe fn wait_for(&mut self, event: &CudaEvent) -> Result<(), DriverError>;
}

impl WaitFor for CudaStream {
    unsafe fn wait_for(&mut self, event: &CudaEvent) -> Result<(), DriverError> {
        self.device().bind_to_thread()?;
        unsafe {
            result::stream::wait_event(
                self.stream,
                event.event(),
                sys::CUevent_wait_flags::CU_EVENT_WAIT_DEFAULT,
            )
        }
    }
}
```
Here, we explicitly state that event recording modifies an event but not the stream, as it merely captures the state of the stream. This makes it harder (though not impossible) to accidentally overwrite an event. However, wait_for submits work to the stream while only reading the event, which correctly results in mutable and immutable borrows, respectively.

I could continue with more examples, but the main point remains: whenever the GPU driver modifies a resource handler, Rust should expose a mutable reference to the corresponding Rust abstraction.

On another note, some redesign of these Rust abstractions should be considered. For example, CudaDevice has no reason to hold stream or event handlers—they should be managed separately:
```rust
pub struct CudaDevice {
....
    pub(crate) stream: sys::CUstream,
    /// Used to synchronize with stream
    pub(crate) event: sys::CUevent,
 .....
}
```
Also, why does it return an Arc? Using multithreading with CUDA requires a lot of additional considerations—just slapping an Arc on it is not enough. So why bother with this anyway if separate abstractions have to be developed to handle multithreading properly?

I have other concerns regarding the overall architecture, but to reiterate: implementing this simple principle—GPU driver modifies a resource handler → Rust makes a mutable borrow; GPU driver reads from a resource handler → Rust makes an immutable borrow—would significantly simplify resource management for users.

coreylowman: > Whenever the GPU driver needs to modify resource handlers, Rust should take mutable references to the corresponding Rust abstractions

Totally agree, and that is what cudarc apis do with the sole exception of Launch - because:
1. On rust side we have no knowledge of what a kernel being launched modifies
2. We can't even verify that people pass in the right number of arguments, or the right length of arguments for the buffer.
That's why that method is marked unsafe. But ideally people should be passing in `&mut slice` as an arg if it needs to modify it.

Another consideration here is the async nature of cuda. Since cuda can modify things behind the scenes due to async nature, there's a possible disconnect between the rust borrow checker and kernels modifying things. With allocation/copy apis we've pretty much made this impossible to do via explicit synchronization where it is needed (Like with copying into host slices). Again we can't make any promises with kernel launching.

> For example, CudaDevice has no reason to hold stream or event handlers—they should be managed separately:

Yep that's exactly what this PR aims to solve, I've removed both in this PR!

> Also, why does it return an Arc? Using multithreading with CUDA requires a lot of additional considerations—just slapping an Arc on it is not enough. So why bother with this anyway if separate abstractions have to be developed to handle multithreading properly?

Originally it was just an Rc, but it was actually relatively easy to do multi threading with cuda (once we added `CudaDevice::bind_to_thread()` calls where needed. There are some libraries (like cublas) that you shouldn't pass between threads, and cudarc does not implement Send & Sync for those as is appropriate.

I'm not really sure what abstractions you think are necessary so you lost me there.

> Also, associating CudaSlice with streams feels wrong, this abstraction should just hold data about the device on which it was allocated, and it is up to the user to figure out a proper synchronization.

The reason I am leaning towards associating (via storing a stream inside CudaSlice object), is for the sole purpose of knowing what stream a cuda slice will be freed on. `cuda::free_async()`. Currently it uses whatever stream the `CudaDevice` was allocated with, which in latest release is just the default stream unless you call `CudaDevice::new_with_stream`.

There is good reason for this - PyTorch also treats allocations as associated with a stream. You can read more about that https://github.com/pytorch/pytorch/blob/main/c10/cuda/CUDACachingAllocator.cpp#L59.

Managing what allocations are managed by what streams is a problem in pytorch world, so much so that the comment i linked mentiosn they have a recordStream method for allocations:
> // The library provides a recordStream() function to help insert the correct
// synchronization when allocations are used on multiple streams. This will
// ensure that the block is not reused before each recorded stream completes
// work.

Another view point: allocations are associated with streams because all work on them has to take place on streams. They are allocated on streams, they are modified on streams, and they are freed on streams. You can't really have an allocation without thinking about streams.

serhii-havrylov: > Totally agree, and that is what cudarc apis do with the sole exception of Launch - because:
I didn't mean the parameters of the kernel but corresponding resource handlers. For example in the current kernel launch implementation you pass immutable reference to cuda stream which is wrong it should be &mut. Async nature of cuda here is irrelevant since the gpu driver queues the work for execution synchronously and changes corresponding cuda resources synchronously.

> The reason I am leaning towards associating (via storing a stream inside CudaSlice object)
This makes sense only for the memory allocated with the cuMemAllocAsync. So it would make sense to have two separate types for blocking and async allocations. I guess even if we don't store the stream with which data was allocated we need to create and store the corresponding event that indicates allocation completion so that we can properly synchronize in drop function.

coreylowman: > The asynchronous nature of CUDA is irrelevant here since the GPU driver queues the work for execution synchronously and modifies the corresponding CUDA resources synchronously.

I agree that it *queues* the work synchronously, but that's where the sync ends. Cuda resources are not modified synchronously with the exception of htod copying from unpinned memory.

With the api on the latest release, you can either pass `&slice` or `&mut slice` to LaunchAsync, that is up to user. However LaunchAsync doesn't actually modify either slice, and neither does the cuda driver, it just notes the device pointer to schedule with the kernel.

With the api I am proposing in this PR, it accepts a `&dyn DeviceRepr`. `CudaSlice`, `&CudaSlice`, and `&mut CudaSlice` all implement DeviceRepr, so you can still pass `&slice`, `&&slice`, `&&mut slice`, though the later two are really ugly and I don't expect people to actually use those.

It is really up to the caller of launch to be aware of what can be modified. THat is why the function is marked as unsafe. You can pass in a `&slice` for a pointer that will be modified on CUDA side, and that is legal. You can also pass `&mut slice` for a pointer that will NOT be modified on cuda (and is actually declared as `const *` on cuda side, and that is totally legal.

I like how the new api `[&dyn DeviceRepr; NUM_ARGS]` is much cleaner than the existing tuple implementations for `LaunchAsync`.

Maybe what we need to do is really make it `[&mut dyn DeviceRepr; NUM_ARGS]` to signify that every argument CAN be modified?

> Therefore, it would be logical to have two separate types for blocking and asynchronous allocations. Instead of storing the stream used for allocation, we can create and store the corresponding event that indicates allocation completion, ensuring proper synchronization in the drop function.

I agree in theory, but I think that API might cross territory into being harder to use. There's a line of how far we should go into compile time guaruntees. E.g. we could probably come up with a way at compile time (via generics on CudaSlice object) to track what slice it's on etc, but it would make it much harder to use.

This is the same reason why slapping a lifetime generic on a struct `CudaSlice<'a>` in theory is what you should do, but in practice people hate working with lifetimes, so instead a lot of people just slap `Rc<RefCell<>>` on things.

I'm not saying I know where the line is, its just something to be aware of.

> we can create and store the corresponding event that indicates allocation completion, ensuring proper synchronization in the drop function.

I do really like this idea, but there will be some cudarc users who absolutely do not want to pay that overhead. So we can maybe think about how to do this by default, but make it possible for people to opt out.

serhii-havrylov: Another point I wanted to make is that there is no way to make the `drop` method safe for device data. The user has to ensure that all necessary synchronization is performed; otherwise, `drop` can lead to undefined behaviuor (UB). I think it would make sense to handle manual memory management in this case. `CudaSlice` would monitor whether freeing its memory was scheduled at some point, and inside the drop method, it would panic if it wasn’t. It’s important to note that checking for scheduling is sufficient; we don’t need to wait for the memory to be actually freed.

The same point applies to dropping `CudaStream`s. Currently, you force synchronization on drop, but querying the stream and panicking would make more sense. Forcing device synchronization can hide issues with user-defined synchronization. For example, imagine a `CudaStream` being dropped while it still has some work scheduled on it. I would prefer a panic that indicates I’m doing something wrong, rather than silently executing something by forcing device synchronization.

coreylowman: We almost need to do something like this:

```rust
pub struct OnStream<T> {
    t: T,
    stream: CudaStream,
}

impl<T> OnStream<T> {
    pub fn into_inner(self) -> T {
        todo!("use event to synchronize with self.stream")
    }

    pub unsafe fn as_ref(&self) -> &T {
        todo!()
    }

    pub unsafe fn as_mut(&mut self) -> &mut T {
        todo!()
    }

    pub fn move_to(self, stream: CudaStream) -> Self {
        todo!("do stream synchronization")
    }

    pub unsafe fn move_to_no_sync(self, stream: CudaStream) -> Self {
        todo!("no stream synchronization")
    }
}
```

and then all the stream allocations would return `OnStream<CudaSlice<T>>`

coreylowman: > Another point I wanted to make is that there is no way to make the drop method safe for device data.

Well this really on happens when multiple streams come into the picture. Current release of cudarc is safe because all allocations & memcpy & drops occur on the same exact stream (whatever stream the device is using). It becomes unsafe if you call LaunchAsync::launch_on_stream, but that is why that method is marked as unsafe.

So that is why this PR is so important to get right, because there's so many easy ways to get this behavior wrong once multiple streams come into play.

But if you're only ever using 1 stream, everything is safe.

> The same point applies to dropping CudaStreams. Currently, you force synchronization on drop, but querying the stream and panicking if there is still some work scheduled would make more sense. ... For example, imagine a CudaStream being dropped while it still has some work scheduled on it.

CudaStreams can be dropped while there is still work, and the work will still be completed. https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__STREAM.html#group__CUDA__STREAM_1g244c8833de4596bcd31a06cdf21ee758

>> In case the device is still doing work in the stream hStream when [cuStreamDestroy()](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__STREAM.html#group__CUDA__STREAM_1g244c8833de4596bcd31a06cdf21ee758) is called, the function will return immediately and the resources associated with hStream will be released automatically once the device has completed all work in hStream.

> Currently, you force synchronization on drop, but querying the stream and panicking if there is still some work scheduled would make more sense.

That is proposed behavior when you've constructed the stream via a `stream.fork` call, which indicates that you want the stream to be synchronized with the original stream. However if you create a stream with `CudaDevice::new_stream()` it won't synchronize to anything on drop.

That is my proposal anyway, not saying it has to be that way.

serhii-havrylov: > > The asynchronous nature of CUDA is irrelevant here since the GPU driver queues the work for execution synchronously and modifies the corresponding CUDA resources synchronously.
> 
> I agree that it _queues_ the work synchronously, but that's where the sync ends. Cuda resources are not modified synchronously with the exception of htod copying from unpinned memory.
It seems like you are still misunderstanding what I am saying. When I say resource handlers, I don't mean device memory; I mean something like a cuBLAS handler or a CUDA stream handler. I am saying that whenever the CUDA driver modifies those handlers, the Rust side should use a &mut reference. This way, whenever I am holding a mutable reference to a stream, I can guarantee that only I can submit work to the stream and that I am the only one responsible for managing everything. Currently, you use regular references in these cases, so there is a situation where I might think that I am the only one responsible for scheduling and managing computation, but in reality, someone else might be submitting work to the stream, and Rust will not say anything about this.

serhii-havrylov: > I do really like this idea, but there will be some cudarc users who absolutely do not want to pay that overhead. So we can maybe think about how to do this by default, but make it possible for people to opt out.

I am not sure what kind of overhead you mean here. One has to use events one way or another to synchronize streams anyway. If a person asynchronously allocates device data, it would probably mean they need to perform some complex orchestration, so it would make sense for a constructor to take a `&mut CudaStream` and a `&mut CudaEvent` as arguments.

coreylowman: > I am not sure what kind of overhead you mean here. One has to use events one way or another to synchronize streams anyway. If a person asynchronously allocates device data, it would probably mean they need to perform some complex orchestration, so it would make sense for a constructor to take a &mut CudaStream and a &mut CudaEvent as arguments.

The event overhead. There are power users who will be able to manually verify that their multi stream code works without events, and they won't want to use events.

coreylowman: We can add event concept, that is easy. I'm not against that at all. I think we are focusing on different parts of the same problem. I'm just trying to say that there are many different users for this library, all with different use cases & needs, and we need to think about all of them. That's great that you want to focus entirely on CudaEvents, and that that makes sense for you. There are many users who won't do that or won't want to. And we need to come up with an API that covers as  many use cases as possible, and still follows safe/unsafe contracts.

This is a draft PR. So nothing here is set in stone.

serhii-havrylov: > The event overhead. There are power users who will be able to manually verify that their multi stream code works without events, and they won't want to use events.

I disagree with this statement. Whenever you have truly multi-stream computation, one has to use events — there is no way around it unless you just have N uni-stream tasks and no synchronization is needed, which is an unrealistic scenario, in my opinion.

aseyboldt: Just jumping in here, because I've been working on a multi-stream application that doesn't use any events.
I'm working on a markov-chain sampler, where we run the same computation several times with different random seeds (chains) to find convergence issues.
Each chain is independent, but since it involves quite a few synchronization points and often doesn't fully use a GPU, it would be good to run those on the same device on different streams.
So I'm really looking forward to a release with this PR. :-)

serhii-havrylov: > Each chain is independent, but since it involves quite a few synchronization points and often doesn't fully use a GPU

It's an example of what I call N independent uni-stream tasks.

> I have no idea though if using events to prevent ordering issues would lead to noticeable overhead for me. Probably not?

The overhead for recording an event and waiting for it on the GPU side is on the order of 1μs. If you take PCIe latency into account, that adds another 1–2μs. However, if you have asynchronous computational time, this host-device communication will be amortized and can be ignored. CUDA events are very lightweight synchronization primitives that provide relatively low-overhead timing and synchronization capabilities.

coreylowman:
Closing this one, see above PR for the new version

PR #338, opened 2025-02-26
make build.rs check if we are building for docs.rs

jonaspleyer:
The [docs.rs/about/builds](https://docs.rs/about/builds) page states that we can check if we are on docs.rs via the following lines:
```rust
if std::env::var("DOCS_RS").is_ok() {
    // ... your code here ...
}
```
This PR implements this and automatically sets the version which is also currently being used as a default cuda version for builds.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -53,6 +53,10 @@ fn main() {
 
 #[allow(unused)]
 fn cuda_version_from_build_system() -> (usize, usize) {
+    if std::env::var("DOCS_RS").is_ok() {
+        return (12, 8);
+    }
+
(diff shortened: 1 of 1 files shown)


jonaspleyer:
I understand the motivation for closing this issue, but this will mean for any dependency with the `cuda_version_from_build_system` a workaround for setting correct features to build on docs.rs is necessary.

coreylowman:
Ah good point, ty for clarifying!

jonaspleyer:
I should have included this as a motivation for the PR in the first place. So sorry for the misunderstanding. :smile:

PR #341, opened 2025-03-06
[Breaking] Add CudaContext, CudaEvent. Refactor CudaStream & launching

### LINKED ISSUE #314: `CudaDevice::htod_copy_into` is unsound
If two `htod_copy_into`s to the same CudaSlice are issued in quick succession, the source for the first copy may be freed before it completes. I haven't been able to find an example that behaves incorrectly because of this, though.

### LINKED ISSUE #340: soundness with async copies
The non-`sync` copy interfaces do not hold onto the host memory, so subsequent modification of the host memory can race with the DMA transfer to device. This is an example with `htod_copy_pinned` #336. Small sizes will spuriously pass this test due to what looks like eager copying, but N=10000 here fails every time in my tests (cuda-12.8, sm_89).
```diff
diff --git i/src/driver/safe/alloc.rs w/src/driver/safe/alloc.rs
index 0271a4a..014a3df 100644
--- i/src/driver/safe/alloc.rs
+++ w/src/driver/safe/alloc.rs
@@ -619,12 +619,17 @@ mod tests {
 
     #[test]
     fn test_htod_copy_pinned() {
-        let truth = [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0];
+        const N: usize = 10000;
+        let truth = [1.0; N];
         let dev = CudaDevice::new(0).unwrap();
-        let mut pinned = unsafe { dev.alloc_pinned::<f32>(10) }.unwrap();
+        let mut pinned = unsafe { dev.alloc_pinned::<f32>(N) }.unwrap();
         pinned.as_mut_slice().clone_from_slice(&truth);
         assert_eq!(pinned.as_slice(), &truth);
         let dst = dev.htod_copy_pinned(&pinned).unwrap();
+        {
+            let p = pinned.as_mut_slice();
+            p[p.len() - 1] += 1.0;
+        }
         let host = dev.dtoh_sync_copy(&dst).unwrap();
         assert_eq!(&host, &truth);
     }
```

I don't want to imply extra synchrony or copying or prevent reusing buffers. One tempting change would be for such copies to return a future/request that retains the borrow (which must wait/observe that the operation completed). The trouble is that one can `mem::forget(request)` to release the lifetime without completing the request, then race as above. I don't have a solution for this other than moving ownership of the host memory into the request, returning it at completion. Unfortunately, that's less ergonomic. That wouldn't be necessary if Rust were to add support for linear types.

coreylowman:
Follow up to #335 

Resolves #340
Resolves #314 

# Breaking API changes

- Add new methods to `trait DeviceSlice`
- Remove `trait LaunchAsync`, see sections below for new launch API.
- Remove implementations of `DeviceRepr` for `&CudaSlice`, `&mut CudaSlice`, `&CudaView`, `&mut CudaViewMut`. These have been replaced with implementations of `AsLaunchParam`.
- Remove `CudaSlice::device()`. Replaced with `CudaSlice::ordinal()` and `CudaSlice::context()`
- `CudaFunction::occupancy_max_active_clusters()` and `CudaFunction::occupancy_max_potential_cluster_size()` now require a `&CudaStream` argument
- Remove field `CudaStream.stream`. Replaced with `CudaStream.cu_stream()`

# TODOs

- [ ] Update & add new documentation
- [x] Add unit tests for multi stream use cases
    - [x] test we can read a slice concurrently from multiple streams
    - [x] test that any writes will properly block multiple streams
- [x] Update examples
- [ ] Do proper stream synchronization in other libraries (cublas/cublaslt/cudnn/curand/nccl)

# Background

Currently there's no way to have an existing CudaDevice use a separate CudaStream for things like allocation/memcpy/memset operations. There's CudaDevice::new_with_stream, but that won't share any loaded kernels that had been previously loaded.

Additionally, the current LaunchAsync trait is a bit too complicated and only supports a small number of compile time arguments.

Further, currently using multiple streams is very unsound/unsafe, because CudaSlices don't do any syncing between streams.

# API Changes

## CudaContext

This is now the lowest level struct that all other structs keep an `Arc` reference to. It has pretty much all of the methods that `CudaDevice` has, except for allocation/etc.

## CudaEvent

CudaEvent is now the main way synchronization between multiple streams is done. Create one using a CudaContext, and you can pretty much do what you expect with a CudaEvent.

- Add `CudaContext::empty_event()`
- Add `CudaStream::record_event()`
- Add `CudaEvent::record()`
- Add `CudaEvent::synchronize()`
- Add `CudaEvent::elapsed_ms()`
- Add `CudaEvent::is_complete()`

## CudaStream

- Add `CudaContext::new_stream()`
- Add `CudaStream::fork()`
- Add `CudaStream::wait()` which lets you wait on an event

### Allocations/Copies

CudaStream now has methods for alloc/memcpy/memset. Eventually we will deprecate using CudaDevice at all.

- Add `CudaStream::null()`
- Add `CudaStream::alloc()`
- Add `CudaStream::alloc_zeros()`
- Add `CudaStream::memset_zeros()`
- Add `CudaStream::memcpy_vtod()` - for copying `Vec<T>` to `CudaSlice<T>`
- Add `CudaStream::memcpy_htod()` - for copying host slices (`&[T]`, `&PinnedHostSlice<T>`, `&Vec<T>`) into existing `CudaSlice<T>`
- Add `CudaStream::memcpy_dtov()` - for copying `CudaSlice<T>` to `Vec<T>`
- Add `CudaStream::memcpy_dtoh()` - for copying `CudaSlice<T>` into host slices (`&mut [T]`, `&mut PinnedHostSlice<T>`, `&mut Vec<T>`)
- Add `CudaStream::memcpy_dtod()` - for copying between `CudaSlice<T>`

### New Kernel launching API

The new api is a builder pattern:

```rust
let f: CudaFunction = ...;
let cfg: LaunchConfig = ...;
let src: CudaSlice<T> = ...;
let mut dst = CudaSlice<T> = ...;
stream.launch_builder(&f)
    .arg(src.len())
    .arg(&src)
    .arg(&mut dst)
    .launch(cfg)
```

- Add `CudaStream::launch_builder(CudaFunction) -> LaunchArgs`
- Add `LaunchArgs::arg()` which pushes an argument (CudaSlice/CudaView/CudaViewMut/rust scalar value)
- Add `LaunchArgs::launch()`
- Add `LaunchArgs::launch_cooperative()`
- Add `trait PushKernelArg` replacing the use of `DeviceRepr` in launching.

Now we can:
1. easily pass any number of arguments without using tuples.
2. Easily specify what stream to launch on

## CudaDevice

This is what CudaDevice looks like now:

```rust
#[derive(Debug)]
pub struct CudaDevice {
    pub(crate) stream: Arc<CudaStream>,
    pub(crate) modules: RwLock<BTreeMap<String, Arc<CudaModule>>>,
}
```

So the difference is that CudaDevice is a stream and a set of loaded modules. After this PR, all `CudaDevice` calls just call `self.stream.<method>()`.

**The future intent is to remove CudaDevice and have the user just hold a `Arc<CudaStream>` instead and manage their `Arc<CudaModule>` themselves.**

## CudaSlice

This is more of an internal change, but figured I'd write about it here. CudaSlice/CudaView/CudaViewMut's now maintain three additional fields:

1. `stream: Arc<CudaStream>` - the stream that it was allocated on, and will be dropped on. Also contains the context it's in.
2. `read: CudaEvent` - an event that records any reads
3. `write: CudaEvent` - an event that records any writes

All stream apis that use CudaSlice/CudaView/CudaViewMut will now do two things:
1. Call either `DevicePtr::block_for_read()` or `DevicePtrMut::block_for_write()` to make sure the stream waits properly for other streams.
2. At the end call either `DevicePtr::record_read()` or `DevicePtrMut::record_write()`.

### DIFF
--- a/examples/01-allocate.rs
+++ b/examples/01-allocate.rs
@@ -1,19 +1,18 @@
-use cudarc::driver::{CudaDevice, CudaSlice, DriverError};
+use cudarc::driver::{CudaContext, CudaSlice, DriverError};
 
 fn main() -> Result<(), DriverError> {
-    let dev = CudaDevice::new(0)?;
+    let ctx = CudaContext::new(0)?;
+    let stream = ctx.default_stream();
 
     // unsafe initialization of unset memory
-    let _: CudaSlice<f32> = unsafe { dev.alloc::<f32>(10) }?;
+    let _: CudaSlice<f32> = unsafe { stream.alloc::<f32>(10) }?;
 
     // this will have memory initialized as 0
-    let _: CudaSlice<f64> = dev.alloc_zeros::<f64>(10)?;
+    let _: CudaSlice<f64> = stream.alloc_zeros::<f64>(10)?;
 
-    // initialize with a rust vec
-    let _: CudaSlice<usize> = dev.htod_copy(vec![0; 10])?;
-
-    // or finially, initialize with a slice. this is synchronous though.
-    let _: CudaSlice<u32> = dev.htod_sync_copy(&[1, 2, 3])?;
+    // initialize with slices!
+    let _: CudaSlice<usize> = stream.memcpy_stod(&vec![0; 10])?;
+    let _: CudaSlice<u32> = stream.memcpy_stod(&[1, 2, 3])?;
 
     Ok(())
 }
--- a/examples/02-copy.rs
+++ b/examples/02-copy.rs
@@ -1,30 +1,29 @@
-use cudarc::driver::{CudaDevice, CudaSlice, DriverError};
+use cudarc::driver::{CudaContext, CudaSlice, DriverError};
 
 fn main() -> Result<(), DriverError> {
-    let dev = CudaDevice::new(0)?;
+    let ctx = CudaContext::new(0)?;
+    let stream = ctx.default_stream();
 
-    let a: CudaSlice<f64> = dev.alloc_zeros::<f64>(10)?;
-    let mut b = dev.alloc_zeros::<f64>(10)?;
+    let a: CudaSlice<f64> = stream.alloc_zeros::<f64>(10)?;
+    let mut b = stream.alloc_zeros::<f64>(10)?;
 
     // you can do device to device copies of course
-    dev.dtod_copy(&a, &mut b)?;
+    stream.memcpy_dtod(&a, &mut b)?;
 
     // but also host to device copys with already allocated buffers
-    dev.htod_copy_into(vec![2.0; 10], &mut b)?;
+    stream.memcpy_htod(&vec![2.0; 10], &mut b)?;
+    // you can use any type of slice
+    stream.memcpy_htod(&[3.0; 10], &mut b)?;
 
-    // if you want to use slices, you can do synchronous copy
-    dev.htod_sync_copy_into(&[3.0; 10], &mut b)?;
-
-    // you can transfer back using reclaim:
-    let mut a_host: Vec<f64> = dev.sync_reclaim(a)?;
+    // you can transfer back using memcpy_dtov
+    let mut a_host: Vec<f64> = stream.memcpy_dtov(&a)?;
     assert_eq!(a_host, [0.0; 10]);
 
-    // or copy back without losing ownership:
-    let b_host = dev.dtoh_sync_copy(&b)?;
+    let b_host = stream.memcpy_dtov(&b)?;
     assert_eq!(b_host, [3.0; 10]);
 
-    // or use a slice
-    dev.dtoh_sync_copy_into(&b, &mut a_host)?;
+    // or transfer into a pre allocated slice
+    stream.memcpy_dtoh(&b, &mut a_host)?;
     assert_eq!(a_host, b_host);
 
     Ok(())
--- a/examples/03-launch-kernel.rs
+++ b/examples/03-launch-kernel.rs
@@ -1,28 +1,34 @@
 use cudarc::{
-    driver::{CudaDevice, DriverError, LaunchAsync, LaunchConfig},
+    driver::{CudaContext, DriverError, LaunchConfig, PushKernelArg},
     nvrtc::Ptx,
 };
 
 fn main() -> Result<(), DriverError> {
-    let dev = CudaDevice::new(0)?;
+    let ctx = CudaContext::new(0)?;
+    let stream = ctx.default_stream();
 
     // You can load a function from a pre-compiled PTX like so:
-    dev.load_ptx(Ptx::from_file("./examples/sin.ptx"), "sin", &["sin_kernel"])?;
+    let module = ctx.load_ptx(Ptx::from_file("./examples/sin.ptx"), &["sin_kernel"])?;
 
     // and then retrieve the function with `get_func`
-    let f = dev.get_func("sin", "sin_kernel").unwrap();
+    let f = module.get_func("sin_kernel").unwrap();
 
     let a_host = [1.0, 2.0, 3.0];
 
-    let a_dev = dev.htod_copy(a_host.into())?;
+    let a_dev = stream.memcpy_stod(&a_host)?;
     let mut b_dev = a_dev.clone();
 
+    // we use a buidler pattern to launch kernels.
     let n = 3;
     let cfg = LaunchConfig::for_num_elems(n);
-    unsafe { f.launch(cfg, (&mut b_dev, &a_dev, n as i32)) }?;
+    let mut launch_args = stream.launch_builder(&f);
+    launch_args.arg(&mut b_dev);
+    launch_args.arg(&a_dev);
+    launch_args.arg(n as i32);
+    unsafe { launch_args.launch(cfg) }?;
 
-    let a_host_2 = dev.sync_reclaim(a_dev)?;
-    let b_host = dev.sync_reclaim(b_dev)?;
+    let a_host_2 = stream.memcpy_dtov(&a_dev)?;
+    let b_host = stream.memcpy_dtov(&b_dev)?;
 
     println!("Found {:?}", b_host);
     println!("Expected {:?}", a_host.map(f32::sin));
--- a/examples/04-streams.rs
+++ b/examples/04-streams.rs
@@ -1,38 +1,41 @@
 use cudarc::{
-    driver::{CudaDevice, DriverError, LaunchAsync, LaunchConfig},
+    driver::{CudaContext, DriverError, LaunchConfig, PushKernelArg},
     nvrtc::Ptx,
 };
 
 fn main() -> Result<(), DriverError> {
-    let dev = CudaDevice::new(0)?;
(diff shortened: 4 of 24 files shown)


serhii-havrylov:
I hate to keep bringing up this point, but simply adding a cuda stream and an event as additional fields for CudaSlice is not enough to guarantee any form of runtime safety. The current implementation of `record_use` is unsafe for several reasons.

CUDA events are lightweight synchronization primitives, and even if an event's status is marked as completed, there is no guarantee that all streams waiting on that event have finished their work and had a chance to check the event status. As a result, calling `record_use` can lead to incorrect synchronization - particularly for "slower" kernels/streams, which may incorrectly perceive the event as not completed (even though it was completed and merely recorded again).

To achieve any meaningful level of runtime safety, you essentially need to construct a computational graph in some form. Only then can you ensure that nodes are properly synchronized. For example, events should be associated with instances of CUDA operations (such as memory copies and kernel executions), and synchronization should be enforced by allowing nodes to be connected only if they were produced by the same stream or are explicitly synchronized via an event.

coreylowman:
> and even if an event's status is marked as completed, there is no guarantee that all streams waiting on that event have finished their work and had a chance to check the event status.

This is not how events & stream waits work. Please read up on the cuda driver api for why the propsed api is safe: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EVENT.html#group__CUDA__EVENT_1g95424d3be52c4eb95d83861b70fb89d1 before you comment again.

Stream waits copy the state of the event at the wait_event call, they don't store a reference to the event. So if i call wait_event, and then after that call event.record, it won't change the previous stream wait at all.

> As a result, calling record_use can lead to incorrect synchronization - particularly for "slower" kernels/streams that are waiting for the CudaSlice event, which may incorrectly perceive the event as not completed (even though it was completed and merely recorded again).

This is incorrect - again consult the driver documentation. Here's an example:

| stream 1 | stream 2 | stream 3 | comment |
| --- | --- | --- | --- |
| wait(A) | | | |
| schedule kernel1(A) | | | |
| record(A) | | | |
| |  wait(A) |  | will make stream2 wait until stream1 finishes the kernel |
| | schedule kernel2(A) | | |
| | record(A) | | will overwrite the event stored on A, but doesn't impact any previous wait calls |
| start kernel1(A) | | | |
| | | wait(A) |  will make stream 3 wait for stream 2 to finish, since that was the most recent record |
| | | schedule kernel3(A) |  |
| | | record(A) | |
| finish kernel1(A) | | | unblocks stream 2 |
| | start kernel2(A) | |  |
| | finish kernel2(A) | | unblocks stream 3 |
| |  | start kernel3(A) | |
| |  | finish kernel3(A) | |

> For example, events should be associated with instances of CUDA operations (such as memory copies and kernel executions)

That's what they are. They are associated with the latest operation that used the data.

---

If you believe the proposed api is unsafe, please provide a concrete example with a test case proving your point.

serhii-havrylov:
> So if i call wait_event, and then after that call event.record, it won't change the previous stream wait at all.

I ran a few tests, and it seems you're right. I was initially confused by the wording in the documentation:  `cuEventRecord can be called multiple times on the same event and will overwrite the previously captured state.` My apologies for the misunderstanding.

serhii-havrylov:
However, I still have to complain :) about using trait objects in the launch function. There is a small overhead associated with them, though it's most likely negligible in this case. That said, we should keep an eye on when [variadic generics](https://github.com/rust-lang/lang-team/blob/master/src/design_notes/variadic_generics.md) become part of the language.

coreylowman:
I'm going to swap it to using a builder for the arguments so we can capture mut vs ref. I realized the current design wouldn't allow for concurrent reads of the same slices. So I'm going to add a read_event & write_event (that is only allocated on write on different stream) to CudaSlice to capture this.

coreylowman:
I realized that if we don't record events for the stream the slices/views were allocated on, then the first use of any separate stream will be unsound, since no work will have been captured.

So unfortunately this means we will always be recording any events, no matter the stream. If someone can figure out how to do this feel free to share.

coreylowman:
Going to update documentation & other libraries in separate PRs

serhii-havrylov:
Hi @coreylowman!

I've noticed that [memcpy_stod](https://github.com/coreylowman/cudarc/blob/9652fdece9ba2024ef55db18b4053179ed69696a/src/driver/safe/alloc.rs#L429) blocks on streams since its record_use [synchronizes](https://github.com/coreylowman/cudarc/blob/9652fdece9ba2024ef55db18b4053179ed69696a/src/driver/safe/host_slice.rs#L65) the stream.

I understand that you likely designed memcpy_stod this way to ensure safety, but this approach feels somewhat restrictive and maybe even counterintuitive - one would typically expect a memcpy operation associated with a stream to be asynchronous. I also recognize that achieving both safety and asynchronicity is challenging.

The use-after-free issue with the source HostSlice could potentially be addressed using lifetime annotations (requiring the source slice to outlive both the destination and the stream). However, this still wouldn't guarantee safety in an asynchronous scenario, as we can't determine precisely when the copy will occur and can't prevent modifications to the source before the copy completes.

One possible solution could be to wrap the slice in an immutable reference wrapper that retains a reference to the source slice, keeping the borrow and preventing modifications. This wrapper could either be manually dropped by the user (for more flexibility) or dropped automatically by rust to reclaim the source. If the copy hasn't finished when the wrapper is dropped, the call could panic, ensuring safety while still allowing flexibility.

What do you think?

coreylowman:
> I've noticed that [memcpy_stod](https://github.com/coreylowman/cudarc/blob/9652fdece9ba2024ef55db18b4053179ed69696a/src/driver/safe/alloc.rs#L429) blocks on streams since its record_use [synchronizes](https://github.com/coreylowman/cudarc/blob/9652fdece9ba2024ef55db18b4053179ed69696a/src/driver/safe/host_slice.rs#L65) the stream.

> I understand that you likely designed memcpy_stod this way to ensure safety, but this approach feels somewhat restrictive and maybe even counterintuitive - one would typically expect a memcpy operation associated with a stream to be asynchronous. I also recognize that achieving both safety and asynchronicity is challenging.

@serhii-havrylov check out the memcpy section of this page: https://docs.nvidia.com/cuda/cuda-driver-api/api-sync-behavior.html#api-sync-behavior

basically anything that isn't the pinned memory will be synchronous wrt host anyway. so the cudarc synchronize call we have is basically a no-op

serhii-havrylov:
> basically anything that isn't the pinned memory will be synchronous wrt host anyway. so the cudarc synchronize call we have is basically a no-op

According to the docs `For transfers between device memory and pageable host memory, the function might be synchronous with respect to host.` This means there's no strict guarantee of synchronicity, and asynchronous memory copies are still possible. In practice, pinned memory is likely to be used for performance reasons, but having asynchronous functionality for Rust vectors would still be a nice addition.

PR #346, opened 2025-03-08
[Breaking] nccl::Comm now constructed with CudaStream

coreylowman:
Follow up to #341 

This is for two main reasons:
1. To align all the libraries to utilize the same API (all of them are moving to construction with streams)
2. To make it clear what stream the Comm object operates on
3. To enable Comm to operate on a separate stream
4. To increase the soundness of multi stream operations with slices/views.

This PR adds ptr.block_for_read/write and ptr.record_read/write calls in all the NCCL apis.

### DIFF
--- a/src/nccl/safe.rs
+++ b/src/nccl/safe.rs
@@ -1,5 +1,5 @@
 use super::{result, sys};
-use crate::driver::{CudaDevice, DevicePtr, DevicePtrMut};
+use crate::driver::{CudaContext, CudaStream, DevicePtr, DevicePtrMut};
 use std::mem::MaybeUninit;
 use std::ptr;
 use std::{sync::Arc, vec, vec::Vec};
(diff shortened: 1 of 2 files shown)


src/nccl/safe.rs
```diff
@@ -224,16 +237,19 @@ impl Comm {
         buff: &mut R,
         peer: i32,
     ) -> Result<result::NcclStatus, result::NcclError> {
-        unsafe {
+        buff.block_for_write(&self.stream).unwrap();
```
coreylowman, line 240: CudaStream::wait() and CudaEvent::record() (which is called below) will not return `Err` unless there are previous async errors.

In order to not have to combine the API errors (NcclError + DriverError), we just call unwrap here.

May be addressed in a future PR (maybe we should combine all errors into a single mega `CudaError`

PR #365, opened 2025-03-27
Adding static linking back in

Narsil:
We are using static linking in https://github.com/huggingface/text-embeddings-inference (in order to reduce binary sizes).

I'm suggesting this change to allow back static linking into cudarc.
As it was when first introduced this shrinks dependencies from a ~500MB to ~200MB for a single cublas pull, it can be more drastic in some instances.

Cheers.

Edit: More complete approach https://github.com/coreylowman/cudarc/pull/366

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ cuda-12060 = []
 cuda-12080 = []
 
 dynamic-linking = []
+static-linking = []
 
 nvrtc = []
 driver = ["nvrtc"]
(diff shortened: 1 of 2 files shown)


coreylowman:
@Narsil have you tested this? I didn't think this actually worked anymore with libloading/bindgen since we generate bindings differently now.

Narsil:
I'm branching off 13.5 instead of `main` here because that's what candle uses at the moment, I can definitely rebase when we have a solution.

PR #366, opened 2025-03-31
[Draft/Proposal] 0.13.5 patch2

Narsil:
This is a fix of https://github.com/coreylowman/cudarc/pull/365 where we revamp the entire way to differentiate the various linking options.

Turns out `main` cudarc NEVER statically links, it always uses libloading which makes all compile time failures silent and become runtime errors, in addition of preventing `static-linking` from working altogether.

This draft is quite opiniated, suggestions are welcome to modify it if necessary (whatever ends up in the branch will be used by TEI in the meantime).

Key factors:
 - Enable `static/dynamic` LINKING (regular linking with `ld`) and dynamic LOADING (with libloading, purely at runtime which simplifies cross compilation)
 - Enforces the use of exclusively one, any combination is considered a bug  (done through build.rs panic).
 - Tries to keep everything consistent and readable in any non-sys code. (result, safe etc..)
 
 
 Proposed solution:
 - Generate both `libloading` (loaded) and "classic" bindings using `bindgen` (linked) with different flags in 2 different categories.
 - Add an `adapter` layer within `loaded` directories to add all functions from the linked directory, to make `lib().#call(...)` which uses the global `Once` library initialization (what used to be there before).
 - Every call of `lib().#call(..)` has been renamed `sys::#call..()` which was what was there before 0.10. I couldn't find any proper way to do the other way around correctly, The idea would be have a dummy calling Struct, but it was much more complex given the output of bindgen code, and the dummy struct would still most likely incur a pointer indirection (where as the opposite is more likely to get inlined, but I haven't actually tested any to confirm those hunches)
 - Use feature flag to import every symbol into `sys::` directly so the rest of the code doesn't need to be version aware.
 
 - Added `bindgen.py` in order to be able to recreate every `sys_xxxx.rs` file directly from a script (I didn't find any good way to do this currently in the repo). There are some simplifications in the script over what's actually OK according to nvidia (nccl versions don't all support cuda11 for instance) but I tried my best to get things working correctly.

Should probably superseed: https://github.com/coreylowman/cudarc/pull/365
Not for future me, magic regexp to update all `lib()` locations: `:%s/lib()\_.\{-}c/sys::c/g`


Narsil:
For code duplication, I thought about it, I'm wondering if there's something already in bindgen, or if we could cook our own, but we could most likely have a code dedup factory that would figure out which functions changed across version and activate them through flags (I expect a large chunk of cuda is mostly additive and never breaking). 

Supporting libloading + static calls in a fused way most likely requires a change in bindgen I think, I expect it to be actually harder then preventing code duplication.

Narsil:
@coreylowman Managed to simplify vastly the generated code by deduping code across versions.
I also rewrote the python script into rust directly (which avoids the necessity to have any particular cli installed aside from `cargo` itself, and libclang for bindgen).

coreylowman:
Will take a look today, exciting to see the number of lines reduced!!

Narsil:
I did another pass.

-> Better UX for the long generation process
-> Removed every file duplication both linked and loaded into a single `sys/mod.rs` and everything is feature gated
-> Remove the feature gates when every version has the symbol
-> The loaded library will fail early when a single symbol is missing (this can easily be changed, given the number of breaking changes across cuda version it's safer that way, but probably less convenient).
-> Removed the function pointer, function member indirection introduced by libloading in bindgen (since we already have the sys::cuXXX indirection).

The code became slightly complex, I'll do another pass to clean this up and add a bit of documentation so you could update it.

Screenshot 2025-04-03 at 23 02 15

This feels satisfying.

coreylowman:
Closing since the other PR was merged

PR #371, opened 2025-04-04
Rebased patch.

Narsil:


Narsil:
https://github.com/coreylowman/cudarc/pull/366

helper/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "cudarc_helper"
```
coreylowman, line 2: This is a nitpick, but thoughts on renaming this binary & the directory to "bindings_generator", or something like that?

Narsil: bindings_generator looks fine to me.

Narsil:
Renamed the binary ! Cheers !

coreylowman:
Woohoo thanks for all this work!

coreylowman:
Did you run unit tests on your machine for the different variants? Getting some failures on mine, wondering if I just have a different setup than yours

Narsil:
I did, what kind of failure.

I had a few issues, which were timing issues (like `driver::safe::launch::tests::test_par_launch` which also failed on my machine on  `main` so I assumed it was normal since it's a timing tests)


I can check it out, I don't think I ran a full matrix sweep of all features though.

Narsil:
The 3 that seems flaky on my end are those:

```
    driver::safe::core::tests::test_pinned_copy_is_faster
    driver::safe::launch::tests::test_multi_stream_concurrent_reads
    driver::safe::launch::tests::test_par_launch
```
They sometime succeed or fail all seem timing related (I'm on a 4xL4 machine atm running 570 driver + cuda 12.8)

PR #387, opened 2025-04-14
Only doing event recording if multiple streams are created

### LINKED ISSUE #386: Only record/wait events once a new stream has been created in a given context
> Interesting that you mention eventRecord/eventSynchronize calls, I do endup seeing quite a bunch of them in my traces, in sequences of cuCtxSetCurrent/cuEventRecord/cuStreamWaitEvent/cuEventDestroy. They seem to be taking a non-trivial amount of time. Is there any way to disable them just to check the impact that they have?
> 
> Image
>  

 _Originally posted by @LaurentMazare in [#368](https://github.com/coreylowman/cudarc/issues/368#issuecomment-2802290325)_

coreylowman:
Resolves #386

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -8,7 +8,8 @@ use std::{
     marker::PhantomData,
     ops::{Bound, RangeBounds},
     string::String,
-    sync::{Arc, RwLock},
+    sync::atomic::{AtomicBool, Ordering},
+    sync::Arc,
     vec::Vec,
 };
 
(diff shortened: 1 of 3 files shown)


coreylowman:
@LaurentMazare if you want to test this branch out, this should remove the extra event records. (unless you are creating multiple streams).

If you are creating multiple streams and still want to be able to disable the event recording and DIY, we could add an environment variable or something 🤔 would have to think about it

LaurentMazare:
Great, that actually makes a huge difference for small models. E.g. on a H100 for a 0.11b model, we go from 233 tok/s to 301 tok/s. This is with a batch size 1 so we certainly expect the code to spend more time scheduling work on the gpu than work actually happening on the gpu but still that's a great improvement.
Would certainly like to have this available on the main releases in one way or another (fwiw I would only care about a single stream here).

coreylowman:
Cool I will add some documentation and then get this merged, should be able to release 0.15.2 later today

LaurentMazare:
Nice, also just a nitpick on the code but rather than a `RwLock<bool>`, using an `AtomicBool` is likely to be slightly more efficient though maybe that doesn't matter too much compared to the complexity of cuda internals.

coreylowman:
Oh great call!! will update that TYSM

PR #390, opened 2025-04-15
Expose cublaslt transc

EricLBuehler:
This (breaking) PR adds `transc` to `MatmulConfig`. While it can be useful to have this exposed, this is a breaking change for what is likely a relatively narrow use case. Please let me know if this makes sense!


EricLBuehler:
Thanks!

PR #391, opened 2025-04-15
Adding a way to disable CudaEvent use tracking for CudaSlice

### LINKED ISSUE #389: Delaying create/destroy for CudaEvent associated with CudaSlice
> So this last change certainly improved things. If it was possible to only create/destroy the read/write cuda events when required, this would likely bring more improvements, but not sure if this could be done in a safe way. 

 _Originally posted by @LaurentMazare in [#368](https://github.com/coreylowman/cudarc/issues/368#issuecomment-2802709469)_

With the changes we've introduced in #387, CudaEvent are now needlessly allocated/destroyed assuming single stream context.

Is it possible to delay the creation of them until they are needed?

It might be possible if we add `RwLock<Option<CudaEvent>>` around them so we can insert them later.

coreylowman:
Resolves #389

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ keywords = ["cuda", "nvidia", "gpu", "nvrtc", "cublas"]
 features = ["cuda-12080", "f16", "cudnn"]
 
 [features]
-default = ["std", "cublas", "cublaslt", "curand", "driver", "runtime", "nvrtc", "dynamic-loading"]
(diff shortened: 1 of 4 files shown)


coreylowman:
FYI @LaurentMazare this was just released on v0.16.0 `unsafe { ctx.disable_event_tracking() }`

LaurentMazare:
Thanks, really cool!

PR #402, opened 2025-04-25
Fix apparent use-after-free in nvrtc.

caelunshun:
It seems like in recent nvrtc versions (at least in 12.8), the program source code and name are not copied in nvrtcCreateProgram, but instead kept as a reference. Previously, cudarc would create a temporary CString which was dropped before destroying the nvrtcProgram. This led to various miscompilations and crashes in nvrtc when compiling many programs in parallel. This commit fixes the problem by keeping the corresponding CStrings alive for the lifetime of the nvrtcProgram.

### DIFF
--- a/src/nvrtc/safe.rs
+++ b/src/nvrtc/safe.rs
@@ -94,12 +94,26 @@ pub fn compile_ptx_with_opts<S: AsRef<str>>(
 
 pub(crate) struct Program {
     prog: sys::nvrtcProgram,
+
+    // Held in this struct to ensure they are not
+    // dropped until after the nvrtcProgram.
+    _src: CString,
+    _name: Option<CString>,
 }
 
 impl Program {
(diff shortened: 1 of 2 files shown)


src/nvrtc/safe.rs
```diff
@@ -94,12 +94,26 @@ pub fn compile_ptx_with_opts<S: AsRef<str>>(
...
+    // dropped until after the nvrtcProgram.
+    _src: CString,
+    _name: Option<CString>,
 }
 
 impl Program {
     pub(crate) fn create<S: AsRef<str>>(src: S, name: Option<&str>) -> Result<Self, CompileError> {
-        let prog = result::create_program(src, name).map_err(CompileError::CreationError)?;
-        Ok(Self { prog })
+        let src = CString::new(src.as_ref().as_bytes())
```
coreylowman, line 106: Can you expand on the addition of `.as_bytes()` here?

PR #412, opened 2025-06-10
safe asum

JonathanWoollett-Light:
Added a safe asum wrapper and refactored the safe directory to support more wrappers.


chelsea0x3b:
Thanks for the feature 👍

PR #413, opened 2025-06-10
Feat: add softmax support

kckeiks:
This PR introduces a safe API for softmax operations (cudnnSoftmaxForward).

cuDNN does not provide a dedicated descriptor for softmax, unlike operations such as activation or convolution. To keep the API consistent, this implementation introduces a Softmax<T> type and a corresponding Cudnn::create_softmax() constructor, following the same design approach used for operations like ActivationDescriptor.


chelsea0x3b:
Ty for the feature!

PR #430, opened 2025-07-18
Add `is_culib_present` function

### LINKED ISSUE #429: A function to check if `culib` will succeed
Currently if `culib` functions fail, they panic via `panic_no_lib_found`.  It'd be convenient to have a function which checks for a particular library and returns a `Result` or an `Option`, so that it can be handled downstream without a panic

kaathewisegit:
A trimmed down version of `culib`, which simply returns `true` if a library was found and `false` otherwise.  It doesn't update the static `LIB` variable from `culib`, because `OnceLock` doesn't allow aborting initialization.  But I think that should be fine, since this function would realistically only be called once on startup.

In theory this function can be made safe by skipping loading the library.  I'm not sure what's the chance of stumbling on `libcuda.so` which isn't actually a dynamic CUDA library.

This is a draft with one such function in `driver`.  If there's an interest in this merging this PR, I'll update all of the modules.

Resolves #429

### DIFF
--- a/bindings_generator/src/merge.rs
+++ b/bindings_generator/src/merge.rs
@@ -315,6 +315,20 @@ impl BindingMerger {
 
                #loading_lib
 
+               pub unsafe fn is_culib_present() -> bool {
+                   let lib_names = [#(#lib_names),*];
+                   let choices = lib_names
+                       .iter()
+                       .map(|l| crate::get_lib_name_candidates(l))
+                       .flatten();
+                   for choice in choices {
+                       if Lib::new(choice).is_ok() {
+                           return true;
+                       }
+                   }
+                   false
+               }
+
                pub unsafe fn culib() -> &'static Lib {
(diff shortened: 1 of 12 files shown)


chelsea0x3b:
Hey I like this idea, feel free to add to other modules! Though this file is currently auto generated via bindings-generator @ https://github.com/coreylowman/cudarc/blob/main/bindings_generator/src/merge.rs#L318, so you can add it to bindings-generator and then rerun (or I can rerun after we merge this)

kaathewisegit:
I've moved it to the bindings generation and re-run the script (that was a lot of libraries!)

It introduced a bunch of formatting-related diffs, though.  Unsure if those are acceptable

chelsea0x3b:
@kaathewisegit ty! I've had the formatting thing happen before, just make sure you run `cargo fmt` - the bindings generator script doesn't run it by default atm. That should reduce the diff by a lot

kaathewisegit:
Huh, indeed!  It's much leaner now

PR #449, opened 2025-08-06
fp8 and fp4 support

zackangelo:
Add dependency and `DeviceRepr` impls for fp8. 

@EricLBuehler currently has these behind a feature in the float8 repo but to avoid circular dependencies this should mirror how half is used in cudarc.

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -747,6 +747,18 @@ unsafe impl DeviceRepr for half::f16 {}
 #[cfg(feature = "f16")]
 unsafe impl DeviceRepr for half::bf16 {}
 
+#[cfg(feature = "f8")]
+unsafe impl DeviceRepr for float8::F8E4M3 {}
+
+#[cfg(feature = "f8")]
+unsafe impl ValidAsZeroBits for float8::F8E4M3 {}
+
+#[cfg(feature = "f8")]
+unsafe impl DeviceRepr for float8::F8E5M2 {}
+
+#[cfg(feature = "f8")]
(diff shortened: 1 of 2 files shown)


chelsea0x3b:
Awesome thanks for this!

src/driver/safe/core.rs
```diff
@@ -747,6 +747,26 @@ unsafe impl DeviceRepr for half::f16 {}
...
+#[cfg(feature = "f8")]
+unsafe impl ValidAsZeroBits for float8::F8E4M3 {}
+
+#[cfg(feature = "f8")]
+unsafe impl DeviceRepr for float8::F8E5M2 {}
+#[cfg(feature = "f8")]
+unsafe impl ValidAsZeroBits for float8::F8E5M2 {}
+
+#[cfg(feature = "f4")]
+unsafe impl DeviceRepr for float4::F4E2M1 {}
```
OlivierDehaene, line 761: I don't think this is actually valid.
As far as I know, `float4::F4E2M1` is backed by a `U8` which would mean in this case that the 4 bits stride between values would not be respected.

@EricLBuehler can you confirm?

OlivierDehaene: Torch avoids the issue by creating a `F4E2M1X2` format but there might be an opportunity do to better?

jorgeantonio21: @OlivierDehaene I opened a PR to add `F4E2M1X2`, following Torch approach. Happy to discuss better ways to do it

See https://github.com/EricLBuehler/float4/pull/2

PR #455, opened 2025-08-24
Remove compile error for unsupported CUDA 13.0 in cudnn module

ikryukov:
Bring back cuDNN support for cuda13 since according to [cuDNN docs](https://docs.nvidia.com/deeplearning/cudnn/backend/latest/reference/support-matrix.html) it is supported now


chelsea0x3b:
Noting that the bindings were updated in #456

PR #468, opened 2025-10-28
Make the `driver` feature not depend on `nvrtc`

johanpel:
The `driver` feature depends on the `nvrtc` feature. Remove this dependency and feature-gate its usages (mostly in tests). This is useful when applications use the driver but don't perform runtime compilation of CUDA code.


chelsea0x3b:
Ty for the pr! easy review

Issue #469, opened 2025-10-29
Windows build broken since 0.17.4


kali:
Since 0.17.4, the windows build si broken.

This was discovered on a native build here:

https://github.com/sonos/tract/actions/runs/18862978567/job/53825115832

And reproduced on a linux cross-compilation with the following:

```
[...]/cudarc% cargo build --target x86_64-pc-windows-gnu   --features  cuda-12090                                                                                                                      tags/v0.17.4
   Compiling cudarc v0.17.4 (/home/kali/dev/github/cudarc)
   Compiling libloading v0.8.8
error[E0277]: the trait bound `CUDA_EXTERNAL_MEMORY_HANDLE_DESC_st: Default` is not satisfied
    --> src/driver/result.rs:1193:15
     |
1193 |             ..Default::default()
     |               ^^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `CUDA_EXTERNAL_MEMORY_HANDLE_DESC_st`

error[E0277]: the trait bound `runtime::sys::cudaExternalMemoryHandleDesc: Default` is not satisfied
   --> src/runtime/result.rs:977:15
    |
977 |             ..Default::default()
    |               ^^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `runtime::sys::cudaExternalMemoryHandleDesc`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `cudarc` (lib) due to 2 previous errors
[1]    3024807 exit 101   cargo build --target x86_64-pc-windows-gnu --features cuda-12090
```

lihongtao1024:
I've encountered this issue too and hope it can be resolved soon.

ozanmuyes:
While waiting for the fix you can change the lines manually;

```diff
// ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cudarc-0.17.4/src/driver/result.rs:1193
- ..Default::default()
+ flags: 0,
+ reserved: [0; 16],
```

Changes will be effective on next `cargo build` and if you want to revert yours and apply authors' fixes just remove parent dependency (for example for me it is the `candle-core` crate) and re-add (e.g. `cargo add candle-core` - again for me it is `candle-core`, yours might differ since this library is a trnsitive dependency)

johanpel:
This is likely due to https://github.com/coreylowman/cudarc/pull/466
Perhaps its best to control this at the `ModuleConfig` level.

[...]

PR #470, opened 2025-10-29
Make Default trait derive module specific

johanpel:
Make the use of bindgen's `derive_default` module-specific. Keep deriving it for anything except the CUPTI module. Fixes: https://github.com/coreylowman/cudarc/issues/469


chelsea0x3b:
Hi, ty for the fast pr - resolving this with #471 instead

PR #475, opened 2025-10-30
Only add typical CUDA paths to the linker search paths if no explicit path is supplied

johanpel:
In the build script `fn link_searches(...)` figures out where CUDA libraries may be.
There are two ways of generating linker search paths
1. supplied through some typical environment variables, e.g. `CUDA_PATH`
2. from some hardcoded "standard locations" that include system paths, e.g. `/usr`.

Currently, the way the linker search paths are generated is that any valid paths generated from 2 are always appended to the linker search paths.

This is a problem when building in Conda environments, because this will cause the linker to see all system-installed libraries. This will cause obscure linker errors as shown below, because Conda environments may use a different `libc` etc. than the system-installed one:

```
  = note: rust-lld: error: undefined symbol: __libc_csu_fini
          >>> referenced by /home/johan/workspace/cudarc/.pixi/envs/cuda-12090/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/../lib/Scrt1.o:(_start)

          rust-lld: error: undefined symbol: __libc_csu_init
          >>> referenced by /home/johan/workspace/cudarc/.pixi/envs/cuda-12090/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/../lib/Scrt1.o:(_start)
```

Therefore, this PR does two things:
1. If any `CUDA_PATH`-like environment variable is defined during building, it will no longer add any "standard locations" to the linker search paths in order to prevent system libraries from being exposed to the linker.
2. For the Conda-specific problem described above, emit a warning at build time when dynamically linking without setting any of the `CUDA_PATH`-like environment variables, just to help out folks running into this scenario.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -251,7 +251,7 @@ fn link_searches(major: usize, minor: usize) -> Vec<PathBuf> {
     // occur. Print a warning with some guidance.
     #[cfg(feature = "dynamic-linking")]
     if env_vars.is_empty() && std::env::var("CONDA_PREFIX").is_ok() {
-        println!("cargo::warning=Detected CONDA_PREFIX in the environment, but no CUDA path was set through one of: {TYPICAL_CUDA_PATH_ENV_VARS:?}. Linker errors are likely to occur. Please ensure the environment contains all required dependencies (e.g. the \"cuda-driver-dev\") and retry building with CUDA_HOME=$CONDA_PREFIX.")
+        println!("cargo::warning=Detected $CONDA_PREFIX, but no CUDA path was set through one of: {TYPICAL_CUDA_PATH_ENV_VARS:?}. Linking to system CUDA libraries; linker errors may occur. To use CUDA installed via conda please ensure the environment contains all required dependencies (e.g. the \"cuda-driver-dev\") and retry building with CUDA_HOME=$CONDA_PREFIX.")
     }
 
     let typical_locations = [


build.rs
```diff
@@ -231,18 +239,22 @@ fn static_linking(major: usize, minor: usize) {
...
+        .filter_map(Result::ok)
+        .collect::<Vec<_>>();
+
+    // When building in a Conda-like environment with dynamic linking, if no
+    // CUDA path is supplied, then it is higly likely that, by defaulting our
+    // linker search paths to the typical locations below, linker errors will
+    // occur. Print a warning with some guidance.
+    #[cfg(feature = "dynamic-linking")]
+    if env_vars.is_empty() && std::env::var("CONDA_PREFIX").is_ok() {
+        println!("cargo::warning=Detected CONDA_PREFIX in the environment, but no CUDA path was set through one of: {TYPICAL_CUDA_PATH_ENV_VARS:?}. Linker errors are likely to occur. Please ensure the environment contains all required dependencies (e.g. the \"cuda-driver-dev\") and retry building with CUDA_HOME=$CONDA_PREFIX.")
```
chelsea0x3b:
```suggestion
        println!("cargo::warning=Detected $CONDA_PREFIX, but no CUDA path was set through one of: {TYPICAL_CUDA_PATH_ENV_VARS:?}. Linking to system CUDA libraries; linker errors may occur. To use CUDA installed via conda please ensure the environment contains all required dependencies (e.g. the \"cuda-driver-dev\") and retry building with CUDA_HOME=$CONDA_PREFIX.")
```

chelsea0x3b:
Great improvement, ty!

PR #479, opened 2025-11-01
Add safe API for querying CUDA function attributes

wizenink:
Adds support for querying CUDA kernel/function properties like register usage, memory requirements, and compilation info.

### Changes
  - **Result layer**: Added `function::get_function_attribute()` wrapper for `cuFuncGetAttribute`
  - **Safe API**: Added `CudaFunction::get_attribute()` and convenience methods:
    - `num_regs()` - registers per thread
    - `shared_size_bytes()` - static shared memory size
    - `const_size_bytes()` - constant memory size
    - `local_size_bytes()` - local memory per thread
    - `max_threads_per_block()` - thread limit
    - `ptx_version()` / `binary_version()` - compilation versions
  - **Example**: Added `10-function-attributes.rs` demonstrating kernel introspection

  ### Usage
  ```rust
  let kernel = module.load_function("my_kernel")?;
  println!("Registers: {}", kernel.num_regs()?);
  println!("Shared mem: {} bytes", kernel.shared_size_bytes()?);
  ```

Useful for kernel optimization, debugging, and runtime capability checking.

### DIFF
--- a/examples/10-function-attributes.rs
+++ b/examples/10-function-attributes.rs
@@ -65,10 +65,7 @@ fn main() -> Result<(), DriverError> {
     println!("=== Optimal Launch Configuration (sin_kernel) ===");
     println!("  Suggested block size:     {}", block_size);
     println!("  Min grid size:            {}", min_grid_size);
-    println!(
-        "  Total threads per grid:   {}",
-        min_grid_size * block_size
-    );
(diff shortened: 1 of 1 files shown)


wizenink:
Should now pass cargo fmt

PR #483, opened 2025-11-02
Add minimal cutensor bindings

wizenink:
This PR adds minimal cutensor support following the existing cudarc patterns:
  - FFI bindings generated with bindgen using `bindgen_generator`
  - Result layer with error handling
  - Safe API with `CuTensor` handle and `TensorDescriptor`
  - Uses `CudaSlice<T>` for tensor data (reusing existing infrastructure)

  ## New safe API

  - **safe layer**:
    - `CuTensor` handle wrapper with version detection
    - `TensorDescriptor` for tensor shape/layout metadata
    -  Does not yet have any data, nor usefulness, WIP   

  ## What's not Included (Yet)

  - Contraction/reduction operation implementations
  -  Examples
  - Tests beyond basic handle creation

  ## Current Status

This is in a very early state,  but I wanted a place to discuss the possible safe API, and check if I'm commiting some war crime against the original design intent :) 

  ## Open Questions for Discussion

  **API design**: Is using `CudaSlice<T>` for tensor data + separate `TensorDescriptor` for metadata
  the right approach?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,7 @@
 //! | [CUPTI](https://docs.nvidia.com/cupti/) | ✅ | ✅ | ✅ |
 //! | [nvtx](https://nvidia.github.io/NVTX/) | ✅ | ✅ | ❌ |
 //! | [cuTENSOR](https://docs.nvidia.com/cuda/cutensor/index.html) | ✅ | ✅ | ❌ |
+//!
 //! CUDA Versions supported
 //! - 11.4-11.8
 //! - 12.0-12.9
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -566,7 +566,14 @@ impl<T> Drop for CudaSlice<T> {
         if let Some(write) = self.write.as_ref() {
             ctx.record_err(self.stream.wait(write));
         }
-        ctx.record_err(unsafe { result::free_async(self.cu_device_ptr, self.stream.cu_stream) });
+        if ctx.has_async_alloc {
+            ctx.record_err(unsafe {
+                result::free_async(self.cu_device_ptr, self.stream.cu_stream)
+            });
+        } else {
+            ctx.record_err(self.stream.synchronize());
+            ctx.record_err(unsafe { result::free_sync(self.cu_device_ptr) });
+        }
     }
 }
 
@@ -1718,7 +1725,7 @@ impl<'a, T> CudaViewMut<'a, T> {
     }
 }
 
-fn to_range(range: impl RangeBounds<usize>, len: usize) -> Option<(usize, usize)> {
+pub(super) fn to_range(range: impl RangeBounds<usize>, len: usize) -> Option<(usize, usize)> {
     let start = match range.start_bound() {
         Bound::Included(&n) => n,
         Bound::Excluded(&n) => n + 1,
--- a/src/driver/safe/mod.rs
+++ b/src/driver/safe/mod.rs
@@ -16,5 +16,5 @@ pub use self::external_memory::{ExternalMemory, MappedBuffer};
 pub use self::graph::CudaGraph;
 pub use self::launch::{LaunchArgs, LaunchConfig, PushKernelArg};
 pub use self::profile::{profiler_start, profiler_stop, Profiler};
-pub use self::unified_memory::UnifiedSlice;
+pub use self::unified_memory::{UnifiedSlice, UnifiedView, UnifiedViewMut};
 pub use crate::driver::result::DriverError;
--- a/src/driver/safe/unified_memory.rs
+++ b/src/driver/safe/unified_memory.rs
@@ -1,11 +1,12 @@
 use core::marker::PhantomData;
+use std::ops::RangeBounds;
 use std::sync::Arc;
 
 use crate::driver::{result, sys};
 
 use super::{
-    CudaContext, CudaEvent, CudaStream, DevicePtr, DevicePtrMut, DeviceRepr, DeviceSlice,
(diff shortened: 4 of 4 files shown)


wizenink:
@coreylowman any hint on if this approach seems correct? ^^

src/lib.rs
```diff
@@ -14,30 +14,30 @@
 //! | [cuFILE](https://docs.nvidia.com/gpudirect-storage/api-reference-guide/index.html#introduction) | ✅ | ✅ | ✅ |
 //! | [CUPTI](https://docs.nvidia.com/cupti/) | ✅ | ✅ | ✅ |
 //! | [nvtx](https://nvidia.github.io/NVTX/) | ✅ | ✅ | ❌ |
-//! 
+//! | [cuTENSOR](https://docs.nvidia.com/cuda/cutensor/index.html) | ✅ | ✅ | ❌ |
```
chelsea0x3b, line 17: Since it seems like based on build script you can staticly link to it?
```suggestion
//! | [cuTENSOR](https://docs.nvidia.com/cuda/cutensor/index.html) | ✅ | ✅ | ✅ |
```

src/cutensor/result.rs
```diff
@@ -0,0 +1,349 @@
...
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct CutensorError(pub sys::cutensorStatus_t);
+
+/// Extension trait to convert cutensorStatus_t to Result
+trait CutensorResult {
+    fn result(self) -> Result<(), CutensorError>;
+}
+
+impl CutensorResult for sys::cutensorStatus_t {
+    fn result(self) -> Result<(), CutensorError> {
```
chelsea0x3b: We don't need a trait for this one, just to keep it consistent with other Result structs
```suggestion
impl sys::cutensorStatus_t {
    pub fn result(self) -> Result<(), CutensorError> {
```

src/cutensor/safe/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+        let version = result::get_version();
+
+        // Validate version compatibility based on CUDA version
+        #[cfg(feature = "cuda-13000")]
+        {
+            if version.0 < 2 || (version.0 == 2 && version.1 < 3) {
+                eprintln!(
+                    "Warning: cuTENSOR {}.{}.{} detected, but CUDA 13.x requires cuTENSOR 2.3+",
+                    version.0, version.1, version.2
+                );
```
chelsea0x3b: Should this be a panic instead?

src/cutensor/safe/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+                feature = "cuda-12080",
+                feature = "cuda-12090"
+            )
+        ))]
+        {
+            if version.0 < 2 {
+                eprintln!(
+                    "Warning: cuTENSOR {}.{}.{} detected, but CUDA 12.x requires cuTENSOR 2.0+",
+                    version.0, version.1, version.2
+                );
```
chelsea0x3b: Similar question - if these are actually incompatible we should panic

src/cutensor/safe/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+/// from being dropped.
+#[derive(Debug)]
+pub struct CuTensor {
+    pub(crate) handle: sys::cutensorHandle_t,
+    pub(crate) stream: Arc<CudaStream>,
+    pub(crate) version: (usize, usize, usize),
+}
+
+unsafe impl Send for CuTensor {}
+unsafe impl Sync for CuTensor {}
```
chelsea0x3b: Can you add a link to CuTensor docstring about thread safety like so:

```
/// # Thread Safety
/// This object is fully thread safe, see here: <TODO insert link> 
```

src/cutensor/safe/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+///     None,              // strides (None = packed/column-major)
+///     sys::cudaDataType_t::CUDA_R_32F,
+///     None,              // alignment (None = 256 bytes)
+/// )?;
+/// ```
+#[derive(Debug)]
+pub struct TensorDescriptor {
+    pub(crate) desc: sys::cutensorTensorDescriptor_t,
+    pub(crate) handle: Arc<CuTensor>,
+}
```
chelsea0x3b: Thoughts on changing to lifetime? We might want to get a more complete example to decide on the safe APIs here.

```suggestion
pub struct TensorDescriptor<'a> {
    pub(crate) desc: sys::cutensorTensorDescriptor_t,
    pub(crate) handle: &'a CuTensor,
}
```

chelsea0x3b:
Re safe APIs: if it ends up taking a bit of design, we can remove the safe API from this PR and just merge the bindings/result level APIs and then come back to safe api in a later PR. Just don't want to merge safe api when its incompletely because ideally we won't break it. Don't currently have a way to signify things are in beta, so atm any changes to safe api are considered breaking.

wizenink:
Maybe leaving the safe api to another PR is the best option for now, will expose result and keep the safe api design in another PR/discussion

src/cutensor/safe/mod.rs
wizenink: Seems reasonable to panic here

src/cutensor/safe/mod.rs
wizenink: Lifetime seems ok to me, but we should make sure that CuTensor is Sync, right?

src/lib.rs
wizenink: Did not test yet with cutensor static because I dont have the binaries here, but I don't see any reason why it shouln't work :)

wizenink:
Resolving all of the conversations for this discussion, will create another PR for the safe API

PR #498, opened 2025-11-26
Initial TRT bindings

wizenink:
This PR aims to add support for TensorRT to cudarc.

As TensorRT is a C++ only lib, bindgen automatic generation does not work here, so a cpp wrapper is created to flatten TensorRT into C code.

Right now, we are not downloading redist files from nvidia, that would be a big TODO.

Working ATM:
* Some basic tests
* TensorRT linked and working with CUDA 13 and CUDA 12

Implementation details:
* TensorRT thin wrapper always statically linked (Very basic wrapper, I dont see the benefit of trying to dynamic load it, and CC only supports static lib generation)
* TensorRT code not being fetched on bindings_generator

Missing:
* Feature gating TRT code in the cpp wrapper depending on the cuda version, maybe passing some defines onto the header coming from rust feature flags
* Expand the trt wrapper to expose more TRT features, this is only a small POC
* Fix some deprecation warnings on the TRT plugin code
* Safe wrappers (This is a big one)

@chelsea0x3b maybe you can check it out and just check if the implementation logic is right or if maybe another path should be taken to implement this

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ features = [
     "nvtx",
     "cupti",
     "cutensor",
+    "tensorrt",
 ]
 
 [features]
@@ -87,6 +88,7 @@ cufile = ["driver"]
 nvtx = ["driver"]
 cupti = ["runtime", "driver"]
 cutensor = ["driver"]
+tensorrt = ["driver"]
 
 std = []
 no-std = ["no-std-compat/std"]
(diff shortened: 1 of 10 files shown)


chelsea0x3b:
@wizenink honestly this looks much cleaner than I expected, I think we could totally move forward with this

wizenink:
Nice :) Will focus on having feature gates synced between rust and the C wrapper for now, and leave  fetching trt redist automatically for the future

chelsea0x3b:
@wizenink I think this really belongs in a separate crate. thank you for the investigation and sorry it took me so long to get back to you.

PR #518, opened 2026-01-16
Add fallback for loading like there is for version

wingertge:
Since `fallback-latest` our users can use any CUDA version, but are still restricted on the other mandatory-but-exclusive feature, the loading features.

This PR adds a new `fallback-dynamic-loading` feature that mostly mirrors `fallback-latest`. It will set `dynamic-loading` *only if no other loading feature is enabled*. This would allow us to build and ship with dynamic loading as a default, while still allowing end users to override it manually.

Just like with the `fallback-latest` issue, the dependency is extremely far down the chain, so handling it via `default-features = false` would be a nightmare for users.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -58,7 +58,7 @@ fn main() {
         not(feature = "dynamic-loading"),
         not(feature = "fallback-dynamic-loading")
     ))]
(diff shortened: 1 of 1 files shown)


chelsea0x3b:
ty! easy to review

PR #520, opened 2026-01-21
fixes a few issues with multi gpu usage in both candle and mistralrs

krampenschiesser:
This pr fixes some issues I encountered when running candle and mistralrs with multiple GPU's.
during layer to device assignment multiple gpu's were initialized and the `dtoh` calls failed.
I added the `bind_to_thread` to the memory operations missing it.
Additionally I implemented the peer copy as an alternative to `dtod`.

### DIFF
--- a/src/driver/result.rs
+++ b/src/driver/result.rs
@@ -1395,6 +1395,20 @@ mod tests {
     }
 
     #[test]
+    fn peer_transfer_self() -> Result<(), DriverError> {
+        let ctx1 = CudaContext::new(0)?;
+        let stream1 = ctx1.default_stream();
+        let a: CudaSlice<f64> = stream1.alloc_zeros::<f64>(10)?;
+
+        let ctx2 = CudaContext::new(0)?;
+        let stream2 = ctx2.default_stream();
+        let b = stream2.clone_dtod(&a)?;
+        let _ = stream1.clone_dtoh(&a)?;
+        let _ = stream2.clone_dtoh(&b)?;
+        Ok(())
+    }
+
+    #[test]
     fn re_associate_context_for_memory_op() -> Result<(), DriverError> {
         let ctx1 = CudaContext::new(0)?;
         if device::get_count()? < 2 {


src/driver/safe/core.rs
```diff
@@ -619,6 +619,26 @@ impl<T> CudaSlice<T> {
     }
 }
 
+impl<T: DeviceRepr + ValidAsZeroBits> CudaSlice<T> {
+    pub fn clone_peer(&self, dst: &Arc<CudaStream>) -> Result<CudaSlice<T>, DriverError> {
+        let dst_slice = dst.alloc_zeros::<T>(self.len())?;
+        let _s1 = SyncOnDrop::sync_stream(&self.stream);
+        let _s2 = SyncOnDrop::sync_stream(dst);
```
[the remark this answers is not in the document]
krampenschiesser: i know this might not be the fastest solution but it matches the syncondrop semantics with events in other places

src/driver/safe/core.rs
```diff
@@ -619,6 +619,26 @@ impl<T> CudaSlice<T> {
...
+            result::memcpy_peer_async(
+                dst.context().cu_ctx,
+                dst_slice.cu_device_ptr,
+                self.context().cu_ctx,
+                self.cu_device_ptr,
+                self.num_bytes(),
+                dst.cu_stream,
+            )?
+        };
+        Ok(dst_slice)
```
chelsea0x3b: Rework:
1. device_ptr/device_ptr_mut for proper event tracking
2. using stream.alloc to prevent extra work of memset zeros
```suggestion
         let mut dst_slice = unsafe { stream.alloc::<T>(self.len()) }?;
        {
            let (dst, _record_dst) = dst_slice.device_ptr_mut(stream);
            let (src, _record_src) = self.device_ptr(stream);
            unsafe {
                result::memcpy_peer_async(
                    stream.ctx.cu_ctx,
                    dst,
                    self.stream.ctx.cu_ctx,
                    src,
                    self.num_bytes(),
                    stream.cu_stream,
                )
            }?;
        }
        Ok(dst_slice)
```

chelsea0x3b: Oh woops forgot I renamed `dst` -> `stream` 😓

src/driver/safe/core.rs
```diff
@@ -619,6 +619,27 @@ impl<T> CudaSlice<T> {
     }
 }
 
+impl<T: DeviceRepr + ValidAsZeroBits> CudaSlice<T> {
+    pub fn clone_peer(&self, dst: &Arc<CudaStream>) -> Result<CudaSlice<T>, DriverError> {
```
chelsea0x3b: sry forgot this line!
```suggestion
    pub fn clone_peer(&self, stream: &Arc<CudaStream>) -> Result<CudaSlice<T>, DriverError> {
```

chelsea0x3b:
@krampenschiesser i think this looks good. as far as usability, i'm wondering if its just better UX to merge this functionality into `stream.clone_dtod`? The inside of the method would just check if the contexts are different and using peer copy if they are

ETA: plus the `stream.<method>` is slightly more clear (IMO) on what stream is actually executing the operation. with `x.clone_peer(stream)` its not exactly clear to me how `stream` is being used or what stream is executing the action

krampenschiesser:
Totally agree, will move it into the stream. I was in the same boat that's why I synchronized both streams initially.

krampenschiesser:
Way cleaner now, stream responsibility is clear and its backwards compatible.
Tests run successfully  once I used the src stream for the event scheduling.

krampenschiesser:
Patched it through to mistral.rs inference works fine with this.

chelsea0x3b:
awesome will merge and cut next release later tonight

krampenschiesser:
Thank you!

PR #523, opened 2026-01-24
Implement `DeviceRepr` for arrays

kaathewisegit:
It's safe because Rust array elements are stored contiguously, without any extra padding:

https://doc.rust-lang.org/reference/type-layout.html#r-layout.array

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -827,6 +827,8 @@ unsafe impl DeviceRepr for float4::E8M0 {}
 #[cfg(feature = "f4")]
 unsafe impl ValidAsZeroBits for float4::E8M0 {}
 
+unsafe impl<const N: usize, T> DeviceRepr for [T; N] where T: DeviceRepr {}
+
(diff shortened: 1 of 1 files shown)


chelsea0x3b:
@kaathewisegit do you know if you can actually pass an array to a cuda kernel as a kernel arg? can you run some tests to see if the cuda kernel can access host memory this way?

kaathewisegit:
You can pass it, but only as a value, not a pointer.  I'm running this PR branch "in production":

https://github.com/kaathewisegit/aspartik/blob/trunk/crates/b3/src/likelihood/cuda/kernels.cu#L158-L169

This includes passing a single `[f64; 4]` (`f64x4` is an alias to CUDA's `double4` here) and a list of arrays in the form of Rust's `Vec<[f64; 4]>`.  The latter is actually the reason why I needed `DeviceRepr` in the first place, since the `Vec<T>` implementation requires `T: DeviceRepr`.

But if someone tries to pass, say `[f64; N]`, and receive it as a `double* arg`, then yes, it would be an error

PR #549, opened 2026-03-22
feat(cublaslt): add safe wrappers for matmul preference and algorithm enumeration APIs

ksutaria:
## Summary

Safe wrappers for cuBLASLt matmul preference control and algorithm enumeration APIs.

Closes #548: the existing `Matmul<T>::matmul()` delegates algorithm selection
entirely to cuBLASLt's heuristic, which silently changes kernel selection based on
available GPU memory. On H100 with FP8 workloads, this produced incorrect results
when memory pressure shifted the heuristic's choice.

This PR adds:

- **`MatmulPreference`** — control heuristic search constraints (e.g., max workspace bytes)
- **`MatmulOperation<T>`** — cuDNN ConvForward-style operation struct with explicit
  algorithm selection: `pick_algorithm()`, `pick_algorithms()`, `get_algo_ids()`,
  `algo_from_id()`, `check_algorithm()`, and `launch()`
- **`MatmulAlgorithm` / `MatmulHeuristicResult`** — safe wrappers for algorithm objects

Three unlocks:
- **Correctness**: constrain workspace so the heuristic cannot silently pick a wrong algorithm
- **Reproducibility**: pin a specific algorithm for deterministic behavior across runs
- **Debuggability**: enumerate and validate all compatible algorithms

The existing `Matmul<T>::matmul()` API is unchanged — this is additive only.

## Test plan

All 5 new tests verified on NVIDIA H100 NVL (CUDA 13.0, Rust 1.94.0):

- [x] `test_matmul_op_pick_algorithm` — pick_algorithm + launch matches existing matmul() output
- [x] `test_matmul_op_pick_algorithms` — returns multiple candidates
- [x] `test_matmul_op_algo_enumeration` — get_algo_ids + algo_from_id + check_algorithm round-trips
- [x] `test_workspace_constrained_correctness` — workspace=0 constraint produces correct results
- [x] `test_pinned_algorithm_reproducibility` — pinned algorithm produces bitwise-identical results

```
running 5 tests
test cublaslt::safe::tests::test_matmul_op_pick_algorithms ... ok
test cublaslt::safe::tests::test_matmul_op_algo_enumeration ... ok
test cublaslt::safe::tests::test_pinned_algorithm_reproducibility ... ok
test cublaslt::safe::tests::test_workspace_constrained_correctness ... ok
test cublaslt::safe::tests::test_matmul_op_pick_algorithm ... ok
```

Note: pre-existing `test_matmul_f32` fails on H100 due to TF32 tolerance (`1e-6` too tight) — not related to this PR.

Clippy and fmt pass clean on our code.

### DIFF
--- a/src/cublaslt/mod.rs
+++ b/src/cublaslt/mod.rs
@@ -15,12 +15,8 @@
 //! 4. Pick an algorithm via [MatmulOperation::pick_algorithm()] or [MatmulOperation::pick_algorithms()]
 //! 5. Execute with [MatmulOperation::launch()]
 //!
-//! For explicit algorithm enumeration:
-//!
-//! 1. Call [MatmulOperation::get_algo_ids()] to list compatible algorithm IDs
-//! 2. Initialize with [MatmulOperation::algo_from_id()]
-//! 3. Validate with [MatmulOperation::check_algorithm()]
-//! 4. Execute with [MatmulOperation::launch()]
+//! Algorithms can also be selected by ID via [MatmulOperation::get_algo_ids()] and
+//! [MatmulOperation::algo_from_id()] for pinning a specific algorithm across runs.
 //!
 //! Note that all above apis work with [crate::driver::DevicePtr]/[crate::driver::DevicePtrMut], so they
 //! accept [crate::driver::CudaSlice], [crate::driver::CudaView], and [crate::driver::CudaViewMut].
--- a/src/cublaslt/mod.rs
+++ b/src/cublaslt/mod.rs
@@ -1,7 +1,26 @@
 //! [CudaBlasLT] wraps around [cuBLASLt](https://docs.nvidia.com/cuda/cublas/index.html#using-the-cublaslt-api) via:
 //!
+//! # Simple path
+//!
 //! 1. Instantiate a [CudaBlasLT] handle with [CudaBlasLT::new()]
 //! 2. Execute a gemm using [CudaBlasLT::matmul()]
+//!
+//! # Advanced path: algorithm selection and preference control
+//!
+//! For control over algorithm selection (e.g., constraining workspace or pinning algorithms):
+//!
+//! 1. Instantiate a [CudaBlasLT] handle with [CudaBlasLT::new()]
+//! 2. Create a [MatmulOperation] with [CudaBlasLT::matmul_op()]
+//! 3. Create a [MatmulPreference] and configure constraints (e.g., [MatmulPreference::set_max_workspace_bytes()])
+//! 4. Pick an algorithm via [MatmulOperation::pick_algorithm()] or [MatmulOperation::pick_algorithms()]
+//! 5. Execute with [MatmulOperation::launch()]
+//!
+//! For explicit algorithm enumeration:
+//!
+//! 1. Call [MatmulOperation::get_algo_ids()] to list compatible algorithm IDs
+//! 2. Initialize with [MatmulOperation::algo_from_id()]
+//! 3. Validate with [MatmulOperation::check_algorithm()]
+//! 4. Execute with [MatmulOperation::launch()]
 //!
 //! Note that all above apis work with [crate::driver::DevicePtr]/[crate::driver::DevicePtrMut], so they
 //! accept [crate::driver::CudaSlice], [crate::driver::CudaView], and [crate::driver::CudaViewMut].
--- a/src/cublaslt/safe.rs
+++ b/src/cublaslt/safe.rs
@@ -8,6 +8,7 @@ use core::ffi::c_int;
 use core::marker::PhantomData;
 use core::mem;
 use std::sync::Arc;
+use std::vec::Vec;
 
 /// Wrapper around [sys::cublasLtHandle_t]
 ///
--- a/src/cublaslt/safe.rs
+++ b/src/cublaslt/safe.rs
@@ -5,6 +5,7 @@ use crate::cublaslt::result::set_matrix_layout_attribute;
 use crate::driver::sys::{CUdevice_attribute, CUdeviceptr};
 use crate::driver::{CudaSlice, CudaStream, DevicePtr, DevicePtrMut, DriverError};
 use core::ffi::c_int;
+use core::marker::PhantomData;
 use core::mem;
 use std::sync::Arc;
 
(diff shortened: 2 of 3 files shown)


src/cublaslt/safe.rs
```diff
@@ -241,34 +303,283 @@ impl Drop for MatmulDesc {
...
         unsafe {
-            // Set workspace size
             result::set_matmul_pref_attribute(
                 self.handle,
                 sys::cublasLtMatmulPreferenceAttributes_t::CUBLASLT_MATMUL_PREF_MAX_WORKSPACE_BYTES,
                 (&size) as *const _ as *const _,
                 mem::size_of::<usize>(),
+            )
+        }
+    }
```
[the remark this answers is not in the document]
ksutaria: Fixed in f8fe41b — changed `usize` to `u64` for both set and get, matching the CUDA `uint64_t` expectation.

src/cublaslt/safe.rs
```diff
@@ -241,34 +303,283 @@ impl Drop for MatmulDesc {
...
+                self.handle,
+                sys::cublasLtMatmulPreferenceAttributes_t::CUBLASLT_MATMUL_PREF_MAX_WORKSPACE_BYTES,
+                (&mut size) as *mut _ as *mut _,
+                mem::size_of::<usize>(),
+                &mut size_written,
             )?;
         }
-        Ok(())
+        Ok(size)
     }
```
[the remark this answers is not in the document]
ksutaria: Applied the `u64` fix in f8fe41b. Declined the `size_written` check — no other attribute getter in the codebase validates `size_written`, and adding it here would be inconsistent with the established pattern. The CUDA driver already returns an error status if the attribute read fails.

src/cublaslt/safe.rs
```diff
@@ -241,34 +304,283 @@ impl Drop for MatmulDesc {
...
+}
+
+/// Result from heuristic algorithm search or algorithm validation.
+#[derive(Debug, Clone, Copy)]
+pub struct MatmulHeuristicResult {
+    /// The selected algorithm with workspace size populated.
+    pub algo: MatmulAlgorithm,
+    /// Estimated number of GPU waves for this algorithm.
+    pub waves_count: f32,
+}
```
chelsea0x3b: Why are these split into 2 separate structs? It's not clear what we gain from this and I think it just complicated the API.

src/cublaslt/safe.rs
```diff
@@ -241,34 +304,283 @@ impl Drop for MatmulDesc {
...
+                max_results,
+            )
+        }
+    }
+
+    /// Initialize an algorithm from a known ID.
+    ///
+    /// The returned `MatmulAlgorithm` has `workspace_size` set to 0.
+    /// Call [`check_algorithm()`](Self::check_algorithm) to get the actual workspace requirement.
+    pub fn algo_from_id(&self, algo_id: c_int) -> Result<MatmulAlgorithm, CublasError> {
```
chelsea0x3b, line 475: It seems like you **must** call check_algorithm on the MatmulAlgorithm returned from this to do anything useful, why not just combine the two functions?

src/cublaslt/mod.rs
```diff
@@ -1,8 +1,27 @@
...
 //! 2. Execute a gemm using [CudaBlasLT::matmul()]
 //!
+//! # Advanced path: algorithm selection and preference control
+//!
+//! For control over algorithm selection (e.g., constraining workspace or pinning algorithms):
+//!
+//! 1. Instantiate a [CudaBlasLT] handle with [CudaBlasLT::new()]
+//! 2. Create a [MatmulOperation] with [CudaBlasLT::matmul_op()]
+//! 3. Create a [MatmulPreference] and configure constraints (e.g., [MatmulPreference::set_max_workspace_bytes()])
+//! 4. Pick an algorithm via [MatmulOperation::pick_algorithm()] or [MatmulOperation::pick_algorithms()]
```
chelsea0x3b, line 15: What is the difference between these two functions? How would a user know which one to pick? Why are there two options here?

src/cublaslt/safe.rs
```diff
@@ -48,6 +50,67 @@ impl Drop for CudaBlasLT {
...
+            c_layout.set_batch(batch_size, stride_c)?;
+        }
+
+        let matmul_desc = MatmulDesc::new(compute_type, scale_type)?;
+        matmul_desc.set_transpose(cfg.transa, Matrix::A)?;
+        matmul_desc.set_transpose(cfg.transb, Matrix::B)?;
+        matmul_desc.set_transpose(cfg.transc, Matrix::C)?;
+
+        Ok(MatmulOperation {
+            handle: self.handle,
```
chelsea0x3b: What are the drop semantics of having this just be a copy of the CublasLTHandle? I think there's nothing preventing a user from dropping the `CudaBlasLT` and this struct would not know.

chelsea0x3b (changes requested):

src/cublaslt/safe.rs
ksutaria: Agreed — the split added complexity without clear benefit. Merged `MatmulHeuristicResult` into `MatmulAlgorithm` and dropped `waves_count` (it's a performance tuning hint, not needed for the correctness/reproducibility goals of this PR). Fixed in d55cd0f.

src/cublaslt/safe.rs
ksutaria: Combined into a single `algo_from_id()` that initializes and validates in one step — returns a fully validated `MatmulAlgorithm` with `workspace_size` populated, or an error if incompatible. Fixed in d55cd0f.

src/cublaslt/mod.rs
ksutaria: Simplified to a single primary workflow (preference → pick → launch). Algorithm enumeration by ID is now a one-line note for users who need to pin a specific algorithm. Fixed in d55cd0f.

src/cublaslt/safe.rs
ksutaria: Good catch. `MatmulOperation` now borrows `&'a CudaBlasLT` instead of copying the raw handle. The borrow checker prevents the user from dropping `CudaBlasLT` while a `MatmulOperation` is alive. Fixed in d55cd0f.

chelsea0x3b (changes requested):
Hey this looks great with your changes, one last thing so I can merge this without it being a breaking version release, can you add a type alias called `MatmulPref` that just points to MatmulPreference? that way its not a breaking pr

src/cublaslt/safe.rs
```diff
@@ -241,34 +303,256 @@ impl Drop for MatmulDesc {
...
+/// Obtained from [`MatmulOperation::pick_algorithm()`],
+/// [`MatmulOperation::pick_algorithms()`], or [`MatmulOperation::algo_from_id()`].
+#[derive(Debug, Clone, Copy)]
+pub struct MatmulAlgorithm {
+    pub(crate) inner: sys::cublasLtMatmulAlgo_t,
+    /// Workspace bytes required by this algorithm.
+    pub workspace_size: usize,
+}
+
+impl MatmulAlgorithm {
```
pubfnbar, line 376: To allow users to save a lookup table of `sys::cublasLtMatmulAlgo_t` values on disk, and then create `MatmulAlgorithm` values from those LUT values, I suggest adding the constructor `pub fn from_raw_parts(algo: sys::cublasLtMatmulAlgo_t, workspace_size: usize) -> Self` for `MatmulAlgorithm`. The binary representation of [`cublasLtMatmulAlgo_t`](https://docs.nvidia.com/cuda/cublas/index.html#cublasltmatmulalgo-t) is stable while using a given cuBLAS library version.

PR #554, opened 2026-03-23
Add CUDA stream priority API

w4nderlust:
## Summary

This adds safe wrappers for the CUDA stream priority API:

- `result::stream::create_with_priority()`: wraps `cuStreamCreateWithPriority`
- `result::stream::get_priority_range()`: wraps `cuCtxGetStreamPriorityRange`
- `CudaContext::new_stream_with_priority()`: safe wrapper with the same lifecycle management as `new_stream()`

## Motivation

Stream priorities are useful when running inference workloads alongside a game engine or other real-time renderer on the same GPU. By assigning different priorities to the compute and rendering streams, the CUDA driver can make better scheduling decisions and reduce frame drops.

The underlying driver functions (`cuStreamCreateWithPriority`, `cuCtxGetStreamPriorityRange`) have been stable since CUDA 5.0 and are widely used, but cudarc doesn't expose them yet.

## Implementation

The safe wrapper (`new_stream_with_priority`) follows the same pattern as the existing `new_stream()`:
- Binds the context to the calling thread
- Increments the stream counter (entering multi-stream mode and disabling single-stream optimizations)
- Creates a non-blocking stream with the requested priority
- Returns an `Arc<CudaStream>` with proper lifecycle management

The `get_priority_range()` function is exposed at the `result` level only, since it is a context-level query that callers will typically use once to validate a priority value before creating a stream.

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -672,12 +672,30 @@ impl CudaContext {
     /// This will swap the calling context to multi stream mode [CudaContext::is_in_multi_stream_mode()].
     /// If the context is not already in multiple stream mode, then this function will also call [CudaContext::synchronize()].
     pub fn new_stream(self: &Arc<Self>) -> Result<Arc<CudaStream>, DriverError> {
+        self.new_stream_with_priority(0)
+    }
+
+    /// Create a new [sys::CUstream_flags::CU_STREAM_NON_BLOCKING] stream with the
+    /// specified priority.
+    ///
+    /// Lower numerical values indicate higher priority. Use
+    /// [`result::stream::get_priority_range`] to query the valid range.
+    ///
+    /// This will swap the calling context to multi stream mode
+    /// [`CudaContext::is_in_multi_stream_mode()`].
+    pub fn new_stream_with_priority(
+        self: &Arc<Self>,
+        priority: i32,
+    ) -> Result<Arc<CudaStream>, DriverError> {
         self.bind_to_thread()?;
         let prev_num_streams = self.num_streams.fetch_add(1, Ordering::Relaxed);
         if prev_num_streams == 0 && self.is_event_tracking() {
             self.synchronize()?;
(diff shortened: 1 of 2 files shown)


w4nderlust:
Addressed both points:

**Return order of `get_priority_range`**: I kept the `(least, greatest)` order to match the CUDA driver API parameter order (`cuCtxGetStreamPriorityRange(leastPriority, greatestPriority)`). Swapping it would be surprising for anyone cross-referencing the CUDA docs. Instead I expanded the doc comment with an example to make the "numerically smaller = higher priority" semantics clear.

**Dedup `new_stream` / `new_stream_with_priority`**: Refactored `new_stream` to delegate to `new_stream_with_priority(0)`. The default CUDA stream priority is 0, so the behavior is identical.

src/driver/safe/core.rs
```diff
@@ -672,12 +672,30 @@ impl CudaContext {
     /// This will swap the calling context to multi stream mode [CudaContext::is_in_multi_stream_mode()].
     /// If the context is not already in multiple stream mode, then this function will also call [CudaContext::synchronize()].
     pub fn new_stream(self: &Arc<Self>) -> Result<Arc<CudaStream>, DriverError> {
+        self.new_stream_with_priority(0)
```
chelsea0x3b: I'd rather have code duplicated since not all devices will have access to create_with_priority. Can make this PR only add a new function instead of changing the existing one?

chelsea0x3b (changes requested):

src/driver/safe/core.rs
w4nderlust: Sure, that's how it was before I adjusted for Gemini :) give me a sec

w4nderlust: Ok done, please see the latest commit

chelsea0x3b:
ty for pr!

PR #555, opened 2026-03-23
Skip event recording and waiting when event tracking is disabled

w4nderlust:
## Summary

Fixes a hang that occurs when `disable_event_tracking()` is called after CudaSlices have already been allocated.

## Problem

When `disable_event_tracking()` is called on a CudaContext, any CudaSlices that were allocated before the call still hold `Some(CudaEvent)` in their `read`/`write` fields. Two things go wrong with these "stale" events:

1. **Drop for CudaSlice** calls `stream.wait(read)` and `stream.wait(write)` on them, which can hang because the events were recorded on a different stream or in a different synchronization context that no longer makes sense.
2. **DevicePtr / DevicePtrMut impls** return `SyncOnDrop::record_event(...)` which calls `cuEventRecord` on these stale events, also causing hangs.

This is a real problem for workloads that start with event tracking enabled (the default), do some setup allocations, and then disable event tracking for a performance-critical hot loop (e.g., CUDA graph capture/replay in ML inference, where events are not needed and add unnecessary overhead).

## Fix

Gate on `is_event_tracking()` in:
- `Drop for CudaSlice`: skip `stream.wait(read/write)` when tracking is disabled
- `DevicePtr` impls for `CudaSlice`, `CudaView`, `CudaViewMut`: return `SyncOnDrop::Sync(None)` instead of recording an event
- `DevicePtrMut` impls for `CudaSlice`, `CudaViewMut`: same

When event tracking is disabled, synchronization is the caller's responsibility anyway (that is the whole point of opting out), so skipping these operations is both correct and consistent with the API contract.

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -776,11 +776,13 @@ unsafe impl<T> Sync for CudaSlice<T> {}
 impl<T> Drop for CudaSlice<T> {
     fn drop(&mut self) {
         let ctx = &self.stream.ctx;
-        if let Some(read) = self.read.as_ref() {
-            ctx.record_err(self.stream.wait(read));
-        }
-        if let Some(write) = self.write.as_ref() {
-            ctx.record_err(self.stream.wait(write));
+        if ctx.is_event_tracking() {
+            if let Some(read) = self.read.as_ref() {
+                ctx.record_err(self.stream.wait(read));
+            }
+            if let Some(write) = self.write.as_ref() {
+                ctx.record_err(self.stream.wait(write));
+            }
         }
         if ctx.has_async_alloc {
             ctx.record_err(unsafe {
@@ -1146,10 +1148,12 @@ impl<T> DevicePtr<T> for CudaSlice<T> {
                 stream.ctx.record_err(stream.wait(write));
             }
         }
-        (
-            self.cu_device_ptr,
-            SyncOnDrop::record_event(&self.read, stream),
-        )
+        let sync = if self.stream.context().is_event_tracking() {
+            SyncOnDrop::record_event(&self.read, stream)
+        } else {
+            SyncOnDrop::Sync(None)
+        };
+        (self.cu_device_ptr, sync)
     }
 }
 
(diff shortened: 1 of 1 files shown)


chelsea0x3b (changes requested):
Can you add a test that shows the representative hang? The event record's won't cause hanging.

> Drop for CudaSlice calls stream.wait(read) and stream.wait(write) on them, which can hang because the events were recorded on a different stream or in a different synchronization context that no longer makes sense.

This also doesn't really make sense to me. Please add an example where this problem actually occurs

w4nderlust:
You're right, apologies for the imprecise explanation. Let me clarify the actual scenario where this is a problem.

The hang occurs specifically during CUDA graph capture. The sequence is:
1. Allocate CudaSlices with event tracking enabled (the default). Events get attached to their read/write fields.
2. Call disable_event_tracking() to avoid overhead during graph capture.
3. Begin graph capture with cuStreamBeginCapture on the stream.
4. Access a pre-allocated slice via device_ptr(). The returned SyncOnDrop, on drop, calls cuEventRecord(stale_event, captured_stream). This records a node into the graph referencing an event that was created outside the capture, causing a hang.

The stream.wait(read/write) in Drop for CudaSlice is the same story. cuStreamWaitEvent on a captured stream with an external event creates an invalid cross-graph dependency.

So the fix is specifically about this: when event tracking has been disabled, typically to prepare for graph capture, do not record or wait on stale events that were attached before disabling.

I'll add a test that reproduces this with graph capture.

w4nderlust:
After further investigation, I'm closing this PR. The fix is not actually needed.

When I originally encountered the issue, I made two changes simultaneously: fixing an RwLock deadlock in my own code, where contiguous() on an already contiguous narrow view shared the same Arc<RwLock<Storage>> and caused copy_inplace_from to deadlock, and this event tracking change in cudarc. Since both went in at the same time, I attributed the fix to both.

I went back and wrote eight different reproduction scenarios: single stream, multi-stream, graph capture, multi-stream graph capture, warmup plus capture plus replay loop, dropping stale slices, and more. I tested them on an RTX 3090 with and without this patch, and all pass either way. I also tested my full inference pipeline, including LLaMA and Qwen3.5 with CUDA graphs and multiple warmup and measured runs, with the patch reverted. Everything works fine. The stale events do not actually cause hangs or errors.

The real fix was the RwLock on my side. Very sorry for wasting your time on this, I genuinely thought I fixed a bug.

PR #556, opened 2026-03-23
Make CudaSlice::cu_device_ptr public

w4nderlust:
## Summary

Makes the `cu_device_ptr` field on `CudaSlice<T>` public (was `pub(crate)`).

## Motivation

There are legitimate cases where a caller needs the raw `CUdeviceptr` to perform a `cuMemcpyDtoDAsync` (or similar driver-level memcpy) directly, without going through the event tracking machinery.

The current way to get the raw pointer is through the `DevicePtr` trait, but that returns a `SyncOnDrop` guard that records events. When event tracking is disabled (via `disable_event_tracking()`), this is unnecessary overhead and the guard serves no purpose. Having direct access to the raw pointer lets callers who have already opted out of event tracking perform raw memcpy operations without the ceremony.

This is consistent with the general cudarc philosophy of providing safe defaults while still allowing advanced users to drop down to the driver level when needed. The field is already `pub(crate)`, so it is used internally without restriction. Making it `pub` just extends the same access to downstream crates.

## Changes

One-line change: `pub(crate) cu_device_ptr` becomes `pub cu_device_ptr` on `CudaSlice<T>`.

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -762,7 +762,20 @@ impl CudaStream {
 /// This object is thread safe.
 #[derive(Debug)]
 pub struct CudaSlice<T> {
-    pub(crate) cu_device_ptr: sys::CUdeviceptr,
+    /// Raw CUDA device pointer.
+    ///
+    /// This is public for advanced use cases like direct `cuMemcpyDtoDAsync` calls
+    /// that bypass cudarc's event tracking.
+    ///
+    /// # Safety
+    ///
+    /// The pointer is owned by the `CudaSlice` and will be freed when it is dropped.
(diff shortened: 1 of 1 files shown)


w4nderlust:
Expanded the doc comment with a `# Safety` section covering lifetime and synchronization responsibilities, as suggested.

chelsea0x3b:
Yeah not doing this. This allows mutation and also you are supposed to go through DevicePtr and DevicePtrMut by design. If you can show a benchmark that shows that this overhead is actually impacting things, then feel free to open an issue where we discuss. The risks of this outweigh the benefits of this by so much.

PR #562, opened 2026-04-17
prefix from env for toolkit discovery

kali:
I hitted this problem when trying to test different toolkit version on a CI. This allow to give priority to a environment specified location to search libraries from.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,6 +182,20 @@ pub(crate) fn panic_no_lib_found<S: std::fmt::Debug>(lib_name: &str, choices: &[
     panic!("Unable to dynamically load the \"{lib_name}\" shared library - searched for library names: {choices:?}. Ensure that `LD_LIBRARY_PATH` has the correct path to the installed library. If the shared library is present on the system under a different name than one of those listed above, please open a GitHub issue.");
 }
 
+/// Returns the candidate filenames cudarc will try to `dlopen` when looking up
+/// a dynamically-loaded CUDA sub-library (e.g. `"cudart"`, `"cublas"`).
+///
+/// By default the candidates are bare filenames (e.g. `libcudart.so`,
+/// `libcudart.so.13`, …) and the OS loader resolves them via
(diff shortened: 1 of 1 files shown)


chelsea0x3b (commented):
Could this also be configured in `build.rs` script via adding to the `link_searches` function? Basically seems like we want to add a custom spot to that list.

I'm also wondering how this intersects with setting `$CUDA_HOME`/`$CUDA_PATH`/`$CUDA_ROOT` etc. Like can you just set one of those to the proper directory and achieve the same result?

kali:
Absolutely open to discussing other designs.

About the build.rs script... it's interesting, even though I think it solved a different problem than mine. I typically build on machines where CUDA is not installed (congrats on making this possible !) by setting the cuda target version. 

Another option that i like: prioritize versioned aware paths: assuming nvidia unix installs are stereotypical (I mostly know about the debian family), start looking for libraries in /usr/local/cuda-X.Y then /usr/local/cuda-X then /usr/local/cuda using the targetted version at build time.

About the CUDA_* variables, if I understand correctly, if we want these to work, it is up to Cudarc to honor them by going to the right place for libraries. Typically they are used by shell scripts to setup LD_LIBRARY_PATH for the executable that will load the libraries. So this is roughly equivalent to what I did, except I used a different var name and did not mess with LD_LIBRARY_PATH but tweaked cudarc library discovery.  If we can figure out what the spec is on CUDA_*, and if that something we can honor in Cudarc, then sure, why not. It feels very discoverable, it's actually something that I tried when I stumbled on the issue.

chelsea0x3b:
We use `libloading` to load libraries which does respect LD_LIBRARY_PATH, so I think this can also be solved via that when we do `dynamic-loading`. So I'm leaning towards not merging this unless we can identify a place where LD_LIBRARY_PATH doesn't cover. Let me know what you find!

chelsea0x3b:
Going to close - reopen with more details based on my last comment if you find something

PR #563, opened 2026-04-26
Replace CUmemAllocationHandleType enum with u32

roykuper13:
replace CUmemAllocationHandleType enum with transparent u32 newtype to allow bitwise OR without UB


chelsea0x3b (changes requested):
So sys/mod.rs is auto generated by bindings_generator. You'll have to make the change in that file.

roykuper13:
> So sys/mod.rs is auto generated by bindings_generator. You'll have to make the change in that file.

done (hopefully) :)

chelsea0x3b:
ty for pr!

PR #564, opened 2026-05-06
Supporting multiple NCCL/CUDNN versions via features

### LINKED ISSUE #433: panic: Expected symbol in library: DlSym { desc: "/lib/x86_64-linux-gnu/libcudnn.so.8: undefined symbol: cudnnAdvVersionCheck" }
When cudarc (with cudnn) is a built for cuda 11.8 and `dynamic-loading`, the library will panic on undefined symbol.

Even though [`version_check`](https://github.com/coreylowman/cudarc/blob/e2befed7d9b0352247d51c73554d8dfdb0996579/src/cudnn/result.rs#L48) is only tested on CUDA 12, the dynamic loading on CUDA 11 still tries to load `cudnnAdvVersionCheck` where it doesn't exist.

chelsea0x3b:
Resolves #433 

Previously we only supported a hardcoded version of cudnn (9) and nccl (2.27). This causes issues when people have different versions, as those libraries (& cutensor) all have separate versioning/release cycle from the cuda toolkit.

This PR Address this via
1. Updating bindings_generator to pull multiple versions of each library & merge them based on library specific versions
2. Adding cudarc feature flags to control the version of the bindings you want

This is **non-breaking**, so `-F nccl` & `-F cudnn` features still exist, they just will default to "latest version available in cudarc".

We also include a way to get the version from the build system, which also works with the existing `-F fallback-latest`:
- `-F nccl-version-from-build-system` which will find the header file and pull the nccl version from that.
- `-F cudnn-version-from-build-system` same thing
- `-F cutensor-version-from-build-system` same thing

If you want to depend on a specific version you can of course specify that via a specific version that you can find the complete list of in `Cargo.toml`.

They are all specified as `<library>-{major_version_number:02}{minor_version_numer:03}`, so `nccl-02030` is nccl 2.30, and `cudnn-09021` is cudnn-9.21


src/cudnn/result.rs
```diff
@@ -47,31 +47,16 @@ pub fn get_cudart_version() -> usize {
...
+    #[cfg(feature = "cudnn-08009")]
+    unsafe {
+        sys::cudnnAdvInferVersionCheck().result()?;
+        sys::cudnnAdvTrainVersionCheck().result()?;
+        sys::cudnnCnnInferVersionCheck().result()?;
+        sys::cudnnCnnTrainVersionCheck().result()?;
+        sys::cudnnOpsInferVersionCheck().result()?;
+        sys::cudnnOpsTrainVersionCheck().result()?;
+    }
+    #[cfg(any(feature = "cudnn-09010", feature = "cudnn-09021"))]
```
chelsea0x3b, line 59: Everyone rejoice

PR #576, opened 2026-05-14
Adding safe nccl::Group api

### LINKED ISSUE #575: NCCL group_start()/group_end() are not event aware
Title. For reference here is an implementation of scatter:
```rust
group_start().expect("ncclGroupStart");
if self.rank == src_rank {
    for (r, shard) in inputs.iter().enumerate() {
        self.send(shard, r as u64);                  <-- records event
    }
}
self.recv(output, src_rank as u64);                   <--- records event
group_end().expect("ncclGroupEnd");         <-- work actually submitted to stream
```

The problem is above, when nccl.send/nccl.receive (or any nccl call) is called, our device_ptr and device_ptr_mut calls record events on streams properly.

However when group_start() was called, these nccl calls aren't actually submitting work. The actual work is submitted after group_end is called. So any stream synchronization/waiting that cudarc does to make sure values aren't read before they are written does not wait for the proper event.

chelsea0x3b:
Resolves #575

### DIFF
--- a/src/nccl/safe.rs
+++ b/src/nccl/safe.rs
@@ -1,5 +1,5 @@
 use super::{result, sys};
-use crate::driver::{CudaContext, CudaStream, DevicePtr, DevicePtrMut};
+use crate::driver::{CudaContext, CudaStream, DevicePtr, DevicePtrMut, SyncOnDrop};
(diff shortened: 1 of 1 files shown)


chelsea0x3b:
Closing bc I need to rework the API - the reference lifetimes make it **really** hard to use with Views

Issue #579, opened 2026-05-27
Allow configuring allocation flags for `PinnedHostSlice`


jordan-wu-97:
### Problem

`CudaContext::alloc_pinned()` currently creates `PinnedHostSlice<T>` allocations with
`CU_MEMHOSTALLOC_WRITECOMBINED` unconditionally:

https://github.com/chelsea0x3b/cudarc/blob/3e5d38b5fe5ec81c934bdc2c7207f181772e307d/src/driver/safe/core.rs#L1405-L1428

Write-combined pinned memory can be useful when host memory is primarily written by
the CPU and transferred to the device. However, it has poor CPU read performance.
CUDA's documentation notes that reading from write-combined host memory on the CPU
is prohibitively slow.

https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g572ca4011bfcb25034888a14d4e035b9

Because pinned buffers can have different host access patterns, selecting
`CU_MEMHOSTALLOC_WRITECOMBINED` unconditionally makes `PinnedHostSlice` unsuitable
for cases where CPU reads matter.

### Proposal

Provide an API that allows callers to select the flags used for pinned host
allocations while preserving existing behavior for current callers.

For example:

```rust
pub unsafe fn alloc_pinned_with_flags<T: DeviceRepr>(
    self: &Arc<Self>,
    len: usize,
    flags: u32,
) -> Result<PinnedHostSlice<T>, DriverError>;
```

The existing method could continue using write-combined memory for characteristic backwards compatibility:

```rust
pub unsafe fn alloc_pinned<T: DeviceRepr>(
    self: &Arc<Self>,
    len: usize,
) -> Result<PinnedHostSlice<T>, DriverError> {
    self.alloc_pinned_with_flags(len, sys::CU_MEMHOSTALLOC_WRITECOMBINED)
}
```

PR #580, opened 2026-05-27
feat(driver): allow selecting pinned host allocation flags

jordan-wu-97:
## Summary

- Add `CudaContext::alloc_pinned_with_flags()` so callers can select CUDA pinned host allocation flags.
- Preserve the existing behavior of `alloc_pinned()` by delegating with `CU_MEMHOSTALLOC_WRITECOMBINED`.
- Add a GPU-backed regression test that compares CPU read performance for default and write-combined pinned allocations.

Addresses #579.

## Motivation

`CU_MEMHOSTALLOC_WRITECOMBINED` is useful for host-written buffers, but CPU reads from write-combined memory can be much slower than reads from default pinned host memory. Callers should be able to select flags according to their host access pattern.

## Testing

- `cargo fmt --check`
- `cargo clippy --no-default-features --features cuda-13020,no-std,cudnn,cublas,cublaslt,nvrtc,driver,curand,nccl,dynamic-loading,cufile,cupti,nvtx,cufft --all-targets -- -D warnings`
- NVIDIA B300 pod, debug test profile: `cargo test --lib --no-default-features --features cuda-12080,std,driver,dynamic-loading test_default_pinned_host_reads_are_faster_than_write_combined -- --nocapture`
  - default pinned host reads: `18.811 ms`
  - write-combined host reads: `1.287 s`
- NVIDIA B300 pod, release test profile: `cargo test --release --lib --no-default-features --features cuda-12080,std,driver,dynamic-loading test_default_pinned_host_reads_are_faster_than_write_combined -- --nocapture`
  - default pinned host reads: `1.483 ms`
  - write-combined host reads: `666.643 ms`

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -1405,6 +1405,11 @@ impl<T> Drop for PinnedHostSlice<T> {
 impl CudaContext {
     /// Allocates page locked host memory with [sys::CU_MEMHOSTALLOC_WRITECOMBINED] flags.
     ///
+    /// Write-combined memory is intended for memory primarily written by the
+    /// host before a device transfer; it has poor CPU read performance. Use
+    /// [CudaContext::alloc_pinned_with_flags()] with `0` for default page
+    /// locked host memory if CPU reads matter.
+    ///
     /// See [cuda docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g572ca4011bfcb25034888a14d4e035b9)
     ///
     /// # Safety
@@ -1417,11 +1422,6 @@ impl CudaContext {
     }
 
     /// Allocates page locked host memory with the specified `flags`.
-    ///
-    /// Pass `0` for default page locked host memory. Use
-    /// [sys::CU_MEMHOSTALLOC_WRITECOMBINED] for memory primarily written by the
-    /// host before a device transfer; write-combined memory has poor CPU read
-    /// performance.
     ///
     /// See [cuda docs](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g572ca4011bfcb25034888a14d4e035b9)
     ///
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -1433,10 +1433,13 @@ impl CudaContext {
         flags: u32,
     ) -> Result<PinnedHostSlice<T>, DriverError> {
         self.bind_to_thread()?;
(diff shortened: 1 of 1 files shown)


src/driver/safe/core.rs
```diff
@@ -1412,12 +1412,28 @@ impl CudaContext {
...
     ) -> Result<PinnedHostSlice<T>, DriverError> {
         self.bind_to_thread()?;
-        let ptr = result::malloc_host(
-            len * std::mem::size_of::<T>(),
-            sys::CU_MEMHOSTALLOC_WRITECOMBINED,
-        )?;
+        let ptr = result::malloc_host(len * std::mem::size_of::<T>(), flags)?;
         let ptr = ptr as *mut T;
         assert!(!ptr.is_null());
         assert!(len * std::mem::size_of::<T>() < isize::MAX as usize);
```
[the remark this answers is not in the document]
jordan-wu-97: Fixed in `671d3d4`. The byte count is now computed with `checked_mul` before calling `malloc_host`, then the existing `isize::MAX` slice-size invariant is asserted on that checked value. I kept this as panic-based validation rather than returning `DriverError(CUDA_ERROR_INVALID_VALUE)`, since this is a Rust slice validity invariant and preserves the existing behavior. Added `test_alloc_pinned_panics_on_size_overflow`; it passes in the CUDA test pod.

src/driver/safe/core.rs
```diff
@@ -1405,22 +1405,41 @@ impl<T> Drop for PinnedHostSlice<T> {
 impl CudaContext {
     /// Allocates page locked host memory with [sys::CU_MEMHOSTALLOC_WRITECOMBINED] flags.
     ///
+    /// Write-combined memory is intended for memory primarily written by the
+    /// host before a device transfer; it has poor CPU read performance. Use
+    /// [CudaContext::alloc_pinned_with_flags()] with `0` for default page
+    /// locked host memory if CPU reads matter.
```
chelsea0x3b: I don't see 0 as an option for flags, where are you seeing this from the nvidia docs?

jordan-wu-97: removed the new comment here, might be more confusing than useful.

"no flag" / "0" is a valid option though, and should be the one chosen for CPU reads.

jordan-wu-97: https://developer.download.nvidia.com/assets/cuda/files/CUDA2.2PinnedMemoryAPIs.pdf

see page 6 

```
If the Flags parameter is 0, cuMemHostAlloc() emulates the pre-CUDA 2.2 function
cuMemAllocHost() except that it may allocate more than 4GB of memory on 64-bit
systems. 
```

Issue #277, opened 2024-07-16
Current destructor strategy yields panic-in-panic


workingjubilee:
Observe the following backtrace, from the tail end of a compute-sanitizer report. If a panic occurs when Rust is inside what is supposed to be a "nounwind" function, then Rust will adopt a more cynical attitude and start doing what it takes to contain the clearly-unsalvageable runtime.

```
thread 'main' panicked at ${HOME}/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cudarc-0.11.7/src/driver/safe/core.rs:252:76:
called `Result::unwrap()` on an `Err` value: DriverError(CUDA_ERROR_LAUNCH_FAILED, "unspecified launch failure")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at ${HOME}/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cudarc-0.11.7/src/driver/safe/core.rs:252:76:
called `Result::unwrap()` on an `Err` value: DriverError(CUDA_ERROR_LAUNCH_FAILED, "unspecified launch failure")
stack backtrace:
   0:     0x55b30644ed72 - std::backtrace_rs::backtrace::libunwind::trace::he4ee80166a02c846
                               at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/../../backtrace/src/backtrace/libunwind.rs:105:5
   1:     0x55b30644ed72 - std::backtrace_rs::backtrace::trace_unsynchronized::h476faccf57e88641
                               at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55b30644ed72 - std::sys_common::backtrace::_print_fmt::h430c922a77e7a59c
                               at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/sys_common/backtrace.rs:68:5
   3:     0x55b30644ed72 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hffecb437d922f988
                               at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x55b30647be4c - core::fmt::rt::Argument::fmt::hf3df69369399bfa9
[...]

PR #585, opened 2026-06-08
nccl: expose `Comm::abort` / `get_async_error` and stop `Comm::drop` from panicking

### LINKED ISSUE #584: nccl: no way to recover from a hung/failed collective (`ncclCommAbort` / `ncclCommGetAsyncError` not exposed)
I'm using the nccl module for multi-GPU training and need to survive a rank dropping out (node failure, ECC error, etc.). The standard NCCL recovery procedure for this is:

1. run a watchdog that periodically calls `ncclCommGetAsyncError` on each comm,
2. when it reports an error or a collective just hangs past a timeout, call `ncclCommAbort` from the watchdog thread to unblock the rank that's stuck inside the collective,
3. tear down the comm and rebuild a fresh one for the surviving ranks.

The current safe API can't express any of this. Comm only aborts in Drop, and `comm_abort` / `ncclCommGetAsyncError` aren't reachable through the safe layer, so a wedged collective just hangs forever with no way to break out.

A couple of things would close the gap:

- `Comm::abort(&self)`: abortable from another thread while a collective is blocked (that's the whole point of `ncclCommAbort`).
- `Comm::get_async_error(&self)`: so a watchdog can detect a failed comm without blocking.
- a raw `ncclComm_t` accessor on `Comm`, same as `CudaStream::cu_stream` etc., for anything not yet wrapped.

While testing this I also hit a secondary bug: once you abort a comm out of band, `Comm::drop` aborts it a second time, `ncclCommAbort` returns non-success, and the expect in the destructor panics. A panic in Drop is bad on its own (it'll abort() the process if it fires during unwinding), and here it fires on the normal recovery path.

I'll follow this issue with a PR. I have the changes above working locally.

grenade:
Fixes #584.

Adds the pieces needed to recover from a hung or failed collective instead of hanging forever:

- `Comm::abort(&self)`: wraps `ncclCommAbort`. Takes `&self` so it can be called from a watchdog thread while another thread is blocked inside a collective, which is how you unblock a stuck rank.
- `Comm::get_async_error(&self)` wraps ncclCommGetAsyncError. Returns `Ok(())` for `ncclSuccess`/`ncclInProgress` and `Err` only for a real error code, so a watchdog can poll comm health without blocking.
- `Comm::comm(&self)` -> `sys::ncclComm_t`: raw handle accessor for driving NCCL APIs not yet wrapped here, in the same spirit as `CudaStream::cu_stream`. Happy to rename this to nccl_comm if you'd preferit to match the cu_* convention more closely.

The second commit fixes a panic in `Comm::drop`. It used `comm_abort(...).expect(...)`, which has the same panic-in-Drop problem that #277 described for the driver destructors (a panic while unwinding turns into a process abort). It also fires on the normal recovery path: once you abort a comm out of band, the destructor aborts it a second time, `ncclCommAbort` returns non-success, and the `expect` panics. Changed it to ignore the result, matching the "don't panic in Drop" approach already used elsewhere in the crate.

Scope is `src/nccl/safe.rs` only; no changes to the sys or result layers (the underlying functions were already wrapped). Builds and clippys clean under nccl,cuda-12060,dynamic-loading.

I left the existing TODO(thenerdstation) about finalize-then-destroy in place since that's a separate question from the panic.

### DIFF
--- a/src/nccl/safe.rs
+++ b/src/nccl/safe.rs
@@ -56,8 +56,13 @@ fn convert_to_nccl_reduce_op(op: &ReduceOp) -> sys::ncclRedOp_t {
 impl Drop for Comm {
     fn drop(&mut self) {
         // TODO(thenerdstation): Shoule we instead do finalize then destory?
+        //
+        // Ignore the abort result rather than `expect`: a `Drop` must not
+        // panic, and the communicator may already have been aborted out of
+        // band (e.g. via `Comm::abort` to unblock a hung collective), in
+        // which case this second abort returns a non-success code.
         unsafe {
-            result::comm_abort(self.comm).expect("Error when aborting Comm.");
+            let _ = result::comm_abort(self.comm);
         }
     }
 }
--- a/src/nccl/safe.rs
+++ b/src/nccl/safe.rs
@@ -26,11 +26,6 @@ pub struct Comm {
     world_size: usize,
 }
 
-// SAFETY: `ncclComm_t` is an opaque handle that NCCL allows to be used from
-// multiple threads. In particular `ncclCommAbort` is explicitly designed to be
-// called from a different thread than one blocked inside a collective, which is
-// the whole point of [`Comm::abort`]. The single-abort invariant is enforced by
-// the `aborted` flag, so sharing a `Comm` across threads is sound.
 unsafe impl Send for Comm {}
 unsafe impl Sync for Comm {}
 
@@ -190,18 +185,22 @@ impl Comm {
     /// Escape hatch for driving NCCL APIs not yet wrapped by this crate.
     /// The handle is only valid for this `Comm`'s lifetime; the returned
     /// pointer must not be used after the `Comm` is dropped.
-    pub fn comm(&self) -> sys::ncclComm_t {
+    pub fn cu_comm(&self) -> sys::ncclComm_t {
         self.comm
     }
 
     /// Abort this communicator (`ncclCommAbort`).
     ///
-    /// Unlike [`Drop`] (which also aborts), this works on a live `&self`
-    /// and — crucially — may be called from a *different* thread than one
-    /// currently blocked inside a collective. That is how NCCL unblocks a
-    /// hung or failed collective so the surviving ranks can resynchronise.
-    /// After abort the communicator is unusable; rebuild a fresh `Comm`
-    /// (e.g. via [`Comm::from_rank`]) to continue.
+    /// `ncclCommAbort` frees the communicator's resources; afterwards the
+    /// `Comm` is unusable and a fresh one must be built (e.g. via
+    /// [`Comm::from_rank`]) to continue.
+    ///
+    /// NCCL only permits one thread to operate a communicator at a time, so
+    /// the caller must ensure no other thread is issuing operations on this
+    /// `Comm` while `abort` runs. Interrupting a thread that is *blocked*
+    /// inside a collective additionally requires the communicator to have been
+    /// created non-blocking (so collectives never block in the first place and
+    /// abort can be issued at any point); see the NCCL fault-tolerance docs.
     ///
     /// Idempotent: aborting an already-aborted `Comm` (or one that will be
     /// aborted again by [`Drop`]) is a no-op returning `Ok(())`, since
--- a/src/nccl/safe.rs
+++ b/src/nccl/safe.rs
@@ -2,17 +2,37 @@ use super::{result, sys};
 use crate::driver::{
     CudaContext, CudaStream, CudaView, CudaViewMut, DevicePtr, DevicePtrMut, SyncOnDrop,
 };
-use std::{mem::MaybeUninit, sync::Arc, vec, vec::Vec};
+use std::{
+    mem::MaybeUninit,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    vec,
+    vec::Vec,
+};
 
 pub use result::{group_end, group_start};
 
 #[derive(Debug)]
 pub struct Comm {
     comm: sys::ncclComm_t,
+    /// Set once `ncclCommAbort` has been issued for `comm` (via [`Comm::abort`]
+    /// or [`Drop`]). `ncclCommAbort` frees the communicator, so it must run at
+    /// most once — this flag is swapped to `true` by whichever caller wins.
+    aborted: AtomicBool,
     stream: Arc<CudaStream>,
     rank: usize,
     world_size: usize,
 }
+
+// SAFETY: `ncclComm_t` is an opaque handle that NCCL allows to be used from
+// multiple threads. In particular `ncclCommAbort` is explicitly designed to be
+// called from a different thread than one blocked inside a collective, which is
+// the whole point of [`Comm::abort`]. The single-abort invariant is enforced by
+// the `aborted` flag, so sharing a `Comm` across threads is sound.
+unsafe impl Send for Comm {}
+unsafe impl Sync for Comm {}
 
 #[derive(Debug, Clone, Copy)]
 pub struct Id {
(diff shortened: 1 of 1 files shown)


src/nccl/safe.rs
```diff
@@ -56,8 +56,13 @@ fn convert_to_nccl_reduce_op(op: &ReduceOp) -> sys::ncclRedOp_t {
...
+        // Ignore the abort result rather than `expect`: a `Drop` must not
+        // panic, and the communicator may already have been aborted out of
+        // band (e.g. via `Comm::abort` to unblock a hung collective), in
+        // which case this second abort returns a non-success code.
         unsafe {
-            result::comm_abort(self.comm).expect("Error when aborting Comm.");
+            let _ = result::comm_abort(self.comm);
         }
     }
 }
```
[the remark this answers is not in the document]
grenade: Fixed in 63327a2. Correct on both counts: `ncclCommAbort` frees the communicator so abort() + the abort in Drop would double-free, and Comm was `!Send + !Sync` so the cross-thread abort wasn't usable in safe Rust anyway.

- added unsafe impl Send + Sync for Comm. NCCL supports using a comm from multiple threads, and ncclCommAbort is meant to be called from another thread while one is blocked in a collective.
- added an aborted: AtomicBool swapped exactly once, so whichever of abort()/Drop wins issues the single ncclCommAbort; abort() is now idempotent.

I used an AtomicBool rather than AtomicPtr<ncclComm>, same exactly-once guarantee without having to load-and-null-check the handle across the ten collective methods, so the diff stays small.

src/nccl/safe.rs
```diff
@@ -157,6 +162,44 @@ impl Comm {
...
+    /// detect a wedged/failed collective without itself blocking.
+    pub fn get_async_error(&self) -> Result<(), result::NcclError> {
+        let mut async_err = sys::ncclResult_t::ncclSuccess;
+        unsafe { sys::ncclCommGetAsyncError(self.comm, &mut async_err) }.result()?;
+        match async_err {
+            // Success, or a still-running non-blocking collective, are both
+            // "not failed" — only a real error code signals a wedged comm.
+            sys::ncclResult_t::ncclSuccess | sys::ncclResult_t::ncclInProgress => Ok(()),
+            other => Err(result::NcclError(other)),
+        }
```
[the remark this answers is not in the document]
grenade: I don't think this one applies. The ncclResult_t::ncclInProgress variant is gated by the nccl version feature, not the CUDA version, both ncclResult_t definitions in sys include it. The #[cfg(not(...))] in result.rs guards a match arm's behaviour (map to InProgress vs. fall through to _ => Err), not the variant's existence, so matching on it unconditionally in get_async_error still compiles.

CI builds nccl alongside each CUDA version and cargo-check is green on cuda-11040 through cuda-11070 on both Linux and Windows, which I believe confirms it, so I think no change needed here.

src/nccl/safe.rs
```diff
@@ -157,6 +185,51 @@ impl Comm {
         self.world_size
     }
 
+    /// The raw `ncclComm_t` handle backing this `Comm`.
+    ///
+    /// Escape hatch for driving NCCL APIs not yet wrapped by this crate.
+    /// The handle is only valid for this `Comm`'s lifetime; the returned
+    /// pointer must not be used after the `Comm` is dropped.
+    pub fn comm(&self) -> sys::ncclComm_t {
```
chelsea0x3b:
```suggestion
    pub fn cu_comm(&self) -> sys::ncclComm_t {
```
For consistency with other methods like this over the crate

src/nccl/safe.rs
```diff
@@ -2,18 +2,38 @@ use super::{result, sys};
...
     stream: Arc<CudaStream>,
     rank: usize,
     world_size: usize,
 }
 
+// SAFETY: `ncclComm_t` is an opaque handle that NCCL allows to be used from
+// multiple threads. In particular `ncclCommAbort` is explicitly designed to be
+// called from a different thread than one blocked inside a collective, which is
+// the whole point of [`Comm::abort`]. The single-abort invariant is enforced by
+// the `aborted` flag, so sharing a `Comm` across threads is sound.
```
chelsea0x3b: Can you provide sources for this? Notably from https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/threadsafety.html which talks about thread safety and states:
> NCCL primitives are generally not thread-safe, however, they are reentrant. Under multi-thread environment, it is not allowed to issue NCCL operations to a single communicator in parallel with multiple threads; it is not safe to issue NCCL operations in parallel to independent communicators located on the same device with multiple threads (see [Using multiple NCCL communicators concurrently](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/communicators.html#multi-thread-concurrent-usage)). If the child communicator shares the resources with the parent communicator (i.e., [ncclConfig_t](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/types.html#ncclconfig) by splitShare), it is not allowed to issue NCCL operations to the child and parent communicators in parallel.

> It is safe to operate a communicator from multiple threads as long as users can guarantee only one thread operates the communicator at a time. However, for any grouped NCCL operations, users need to ensure only one thread issues all the operations in the group.

chelsea0x3b (commented):
Please clarify more about thread safety of ncclComm_t

src/nccl/safe.rs
grenade: You're correct, and my SAFETY/abort wording overshot. Thank you for the pointer to threadsafety.html. The accurate contract is the one you quoted: a communicator may be used from multiple threads only one at a time, and per the Fault Tolerance section (https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/communicators.html), aborting safely requires the comm to be created nonblocking and no thread to be inside an NCCL op when ncclCommAbort is called (in nonblocking mode every call except abort/destroy returns immediately, so abort can be issued at any point). My "abort a thread that's blocked inside a collective" wording was the blocking-mode model and is innacurate. I've dropped it.

To be clear about what Send/Sync is and isn't claiming here: it governs moving/sharing the handle across threads, not whether concurrent NCCL operations are valid. ncclComm_t is the same kind of opaque handle as CUstream, cublasHandle_t, etc., and the crate already marks all of those unsafe impl Send + Sync (CudaStream, CudaContext, CudaBlas, DnHandle, CudaFft, ...), leaving operation-ordering correctness to the caller. Marking Comm the same way just lets it be moved into / shared with another thread (e.g. behind an Arc) like CudaStream already can be; the "one thread per comm at a time" rule remains the caller's responsibility, exactly as elsewhere in the crate. The aborted: AtomicBool is the one thing that genuinely needs to be atomic, since ncclCommAbort frees the comm and must run exactly once.

Concretely I've: (1) dropped the inaccurate SAFETY prose so the unsafe impls match the bare style used for the other handle types, (2) rewritten the abort() docs to state the real contract — caller ensures no other thread is operating the comm, and interrupting a blocked op requires a nonblocking comm. Exposing nonblocking creation (via the existing comm_init_rank_config) is the natural follow-up that makes the full watchdog pattern usable; abort + get_async_error are the primitives it builds on.

src/nccl/safe.rs
grenade: Makes sense, have renamed to cu_comm() for consistency.

src/nccl/safe.rs
```diff
@@ -2,18 +2,33 @@ use super::{result, sys};
...
+    /// or [`Drop`]). `ncclCommAbort` frees the communicator, so it must run at
+    /// most once — this flag is swapped to `true` by whichever caller wins.
+    aborted: AtomicBool,
     stream: Arc<CudaStream>,
     rank: usize,
     world_size: usize,
 }
 
+unsafe impl Send for Comm {}
+unsafe impl Sync for Comm {}
```
chelsea0x3b, line 30: Based on the nvidia nccl docs, a comm is not thread safe based on my read. This doesn't seem related to the description (https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/threadsafety.html)

src/nccl/safe.rs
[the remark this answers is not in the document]
chelsea0x3b: @grenade where in this page https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/threadsafety.html do you gather that it is supposed to be safe to use a comm from multiple threads?

PR #588, opened 2026-06-18
bindings_generator: make generation resumable and resilient

corrm:
Split out from #582 at the maintainer's request, so the CUDA 13.3 bump stays minimal and these generator improvements can be reviewed on their own. Independent of any specific CUDA version.

- **Resume support**: skip already-generated bindings so a partial run can pick up where it left off instead of starting over.
- **Skip unnecessary downloads**: when a module's bindings already exist and no pending dependent still needs its headers, resolve the extract directory without downloading the archive (`expected_archive_dir`).
- **Retry transient download failures** with backoff instead of panicking on the first error.
- **Descriptive errors** instead of `unwrap()` when a module is missing `linux-x86_64` `relative_path`/`sha256` in a redist manifest.
- **--cuda-version safety**: skip libs that map to a CUDA version outside the current run instead of indexing a missing primary archive.

Generating from scratch still requires doing all versions once (the merge needs every version's bindings), but reruns after a failure are now cheap.

`cargo build`, `cargo fmt --check` and `cargo clippy` on `bindings_generator` are clean (the two pre-existing `merge.rs` clippy warnings are untouched by this PR).


corrm:
Both findings addressed in fa00cb6's follow-up commit:

- `resolve_archive`: `Path::new(relative_path).file_name().unwrap()` → `with_context`. Good catch — `relative_path` comes from the redist manifest, so this is exactly the kind of manifest-derived unwrap this PR is meant to remove.
- `run_bindgen`: `outfilename.parent().unwrap()` → `with_context`. Can't actually trigger given how the path is built, but it's consistent with the rest of the change.

Note: the sibling `get_cuda_major_archive` has the same `relative_path`/`file_name` unwraps, but those are pre-existing on `main` and outside this PR's diff, so I left them untouched to keep the change scoped. Happy to clean them up in a separate pass if you'd prefer.

`cargo build`, `cargo fmt --check` and `cargo clippy` on `bindings_generator` are clean (the two remaining clippy warnings are pre-existing in `merge.rs`).

PR #592, opened 2026-07-29
Added some missing cupti headers for new cuda versions

amotzop:
There are some new APIs inside `cupti` for newer CUDA versions. I added the relevant headers to `cupti/wrapper.h`.
Some of the headers use C++ header files, which are less supported in bindgen, so I left them commented out.
I also tried to add only the minimal set of headers needed.


chelsea0x3b:
ty! 🚀

Issue #536, opened 2026-02-28
Memory pool support (cuMemPool*)


OneThing98:
I noticed cudarc doesn't have any wrappers for the CUDA memory pool
API (cuMemPoolCreate, cuMemAllocFromPoolAsync, etc.) beyond what's
already in the sys bindings.
From what I can know, cudarc uses cuMemAllocAsync internally for
CudaSlice allocation, which goes through the default pool. But there
doesn't seem to be a way to create custom pools, allocate from a
specific pool, or do things like trimming unused memory.

I came across this while looking at #514 and the downstream candle PR
(huggingface/candle#3352) both deal with memory needing to persist
across CUDA graph replays, and it seems like custom pools are how
CUDA expects you to handle that.

I was thinking this could be approached like this:
1. result-level wrappers in a new pub mod mem_pool for the core
functions: create, destroy, trim_to, get/set attribute, and the
device-level get_default_mem_pool/get_mem_pool/set_mem_pool.

2. A safe-level CudaMemPool type with Drop, plus something like
CudaStream::alloc_from_pool() and CudaContext::default_mem_pool().

Though I'm not sure about a few things and would appreciate guidance:
Should pool-allocated CudaSlices track which pool they came from?
The pool struct (CUmemPoolProps) changed between 11.x and 12.x. So do we target a specific version?
I am not too sure how this should interact with CudaGraph capture..?

Happy to start with the result-level wrappers if this seems like a
reasonable direction.

Issue #501, opened 2025-12-01
stream-capture-safe API


wazi893:
Current kernel launch API is not CUDA Graph capture safe; need a ‘no hidden streams, no hidden sync’ mode

chelsea0x3b:
Can you give some code example or link to documentation that describes what determines if a cuda graph is capture safe?

There's already a way to turn off automatic stream synchronization ([`CudaContext::disable_event_tracking()`](https://docs.rs/cudarc/latest/cudarc/driver/safe/struct.CudaContext.html#method.disable_event_tracking)).

PR #594, opened 2026-08-20
feat(driver): add capture-scoped CUDA graph memory pool

lucifer1004:
Related: #501, #536.

Allocations issued with `cuMemAllocAsync` while a stream is capturing become allocation/free *nodes inside the graph*, and their addresses are re-baked at every instantiation. Workloads that capture once per shape at startup and then replay thousands of times (e.g. inference decode loops) instead want all capture-time allocations to land in one preallocated arena at stable addresses, keeping graphs free of alloc/free nodes entirely.

This PR adds `CudaGraphMemoryPool`, a device-local arena shared by graphs that are never replayed concurrently:

- `CudaStream::probe_graph_memory_pool()` measures the aligned storage a warmup allocation sequence needs, without changing its behavior.
- `CudaGraphMemoryPool::new(stream, capacity)` allocates the arena.
- `unsafe begin_session()` starts a bump-allocation session: every `CudaStream::alloc`/`null` on that stream is routed into the arena at 256-byte alignment, and the resulting `CudaSlice`s are pool-owned (their `Drop` does not free). `CudaGraphMemoryPoolSession::finish()` reports the bytes consumed.

Session state is thread-local, so unrelated allocations in the same process are unaffected. The TLS-based session API is gated off under `no-std` (`no_std_compat` has no `thread_local!`); the pool type itself remains available there. Happy to restructure this (e.g. stream-owned session state) if you prefer.

`CudaSlice::leak()` now refuses to transfer ownership of pool-backed slices, since the arena — not the caller — owns those addresses.

Verified downstream: a DeepSeek V4 TP2 inference engine captures its 43-layer decode graph against one arena per rank, removing 105 alloc + 105 free nodes per rank.

### DIFF
--- a/src/driver/safe/core.rs
+++ b/src/driver/safe/core.rs
@@ -1,5 +1,6 @@
 use crate::driver::{
     result::{self, DriverError},
+    safe::graph_memory::{route_allocation, CudaGraphMemoryPoolInner},
     sys::{self, CUfunc_cache_enum, CUfunction_attribute_enum},
 };
 
@@ -791,6 +792,8 @@ pub struct CudaSlice<T> {
     pub(crate) read: Option<CudaEvent>,
     pub(crate) write: Option<CudaEvent>,
     pub(crate) stream: Arc<CudaStream>,
+    pub(crate) graph_memory: Option<Arc<CudaGraphMemoryPoolInner>>,
+    pub(crate) external_owner: Option<Arc<dyn std::fmt::Debug + Send + Sync>>,
     pub(crate) marker: PhantomData<*const T>,
 }
 
@@ -799,6 +802,9 @@ unsafe impl<T> Sync for CudaSlice<T> {}
 
 impl<T> Drop for CudaSlice<T> {
     fn drop(&mut self) {
+        if self.graph_memory.is_some() || self.external_owner.is_some() {
+            return;
+        }
         let ctx = &self.stream.ctx;
         if ctx.is_managing_stream_synchronization() {
             if let Some(read) = self.read.as_ref() {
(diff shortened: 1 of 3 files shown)


src/driver/safe/core.rs
```diff
@@ -791,6 +792,8 @@ pub struct CudaSlice<T> {
     pub(crate) read: Option<CudaEvent>,
     pub(crate) write: Option<CudaEvent>,
     pub(crate) stream: Arc<CudaStream>,
+    pub(crate) graph_memory: Option<Arc<CudaGraphMemoryPoolInner>>,
+    pub(crate) external_owner: Option<Arc<dyn std::fmt::Debug + Send + Sync>>,
```
chelsea0x3b, line 796: I'm not sure if this design is the correct way forward. CudaSlice is almost turning into an enum at this point:
```rust
enum CudaSlice<T> {
    Owned( ... ),
    External(...),
    PoolAllocated(...),
    ...,
}
```

We should probaly discuss this a bit more. I definitely want to support memory pools

lucifer1004: Agreed, this is the wrong shape. Looking at it again, it also bundles something the description doesn't mention: `upgrade_external_device_ptr` / `external_owner` (for VMM and IPC memory). That's how `CudaSlice` ended up with two separate "someone else frees this" fields.

Also, a correction to the description: the downstream engine I cited doesn't depend on this branch right now, and "105 alloc + 105 free nodes removed" is a node count, not a measured latency win. Please don't weigh the PR on that.

Here's a smaller shape I'd propose instead:

- **cudarc gets one primitive:** a single optional keep-alive owner on `CudaSlice` (`Option<Arc<dyn Any + Send + Sync>>`) plus one constructor, e.g. `CudaStream::upgrade_device_ptr_with_owner(ptr, len, owner)`. When the owner is set, `Drop` releases the owner and does not free the pointer, and `leak` refuses. Arena sub-allocations, VMM mappings and IPC imports all use this same path, so `CudaSlice` doesn't grow a new variant for each allocation source.
- **Pool/arena policy stays out of cudarc for now:** no thread-local allocation routing, no probe/session. Downstream can build a bump arena on the primitive above.
- **Real memory-pool support** (`cuMemPoolCreate` / `cuMemAllocFromPoolAsync` / trim / attributes) follows #536 as a separate PR. Pool-allocated slices could hold their pool through the same owner field.

If that direction works for you, I'll close this and open the primitive as a small PR. Or tell me if you'd rather have pools in `CudaSlice` some other way.

PR #595, opened 2026-08-20
feat(cublas): support a retained caller-owned workspace

lucifer1004:
During CUDA Graph capture, cuBLAS operations that need workspace record internal `cuMemAllocAsync`/`FreeAsync` nodes into the graph (see the cuBLAS docs on `cublasSetWorkspace`), unless the caller assigns a workspace before capture begins.

This PR wraps `cublasSetWorkspace_v2` at the result layer and adds `CudaBlas::set_workspace`, which binds and retains a caller-owned `CudaSlice<u8>` for the handle's lifetime. `set_stream` re-binds the retained workspace after a stream change. The binding validates that the workspace belongs to the handle's context.

Same motivation as the graph memory pool PR: keeping captured graphs free of library-internal allocation nodes.

### DIFF
--- a/src/cublas/safe/mod.rs
+++ b/src/cublas/safe/mod.rs
@@ -2,8 +2,8 @@
 #![allow(clippy::too_many_arguments)]
 
 use super::{result, result::CublasError, sys};
-use crate::driver::CudaStream;
-use std::sync::Arc;
+use crate::driver::{CudaSlice, CudaStream, DevicePtrMut};
+use std::sync::{Arc, Mutex};
 
 mod asum;
 mod gemm;
@@ -27,6 +27,7 @@ pub use grouped_gemm::*;
 pub struct CudaBlas {
     pub(crate) handle: sys::cublasHandle_t,
     pub(crate) stream: Arc<CudaStream>,
+    workspace: Mutex<Option<CudaSlice<u8>>>,
 }
 
 unsafe impl Send for CudaBlas {}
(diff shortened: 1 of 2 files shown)


src/cublas/safe/mod.rs
```diff
@@ -56,7 +61,51 @@ impl CudaBlas {
...
+    pub unsafe fn set_workspace(&self, mut workspace: CudaSlice<u8>) -> Result<(), CublasError> {
+        self.bind_workspace(&mut workspace)?;
+        *self
+            .workspace
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner) = Some(workspace);
+        Ok(())
+    }
+
+    fn bind_workspace(&self, workspace: &mut CudaSlice<u8>) -> Result<(), CublasError> {
```
chelsea0x3b, line 92: Can you explain why this is needed?

lucifer1004: It's there for two reasons, both measured (CUDA 13.3, cuBLAS 13, one 4096³ SGEMM captured per case):

1. **`set_stream` has to re-bind.** Per the [`cublasSetStream` docs](https://docs.nvidia.com/cuda/cublas/#cublassetstream), setting a stream *unconditionally resets the workspace to cuBLAS's default pool*. So a workspace assigned once is silently dropped on the first `set_stream`, and the graph gets allocation nodes again:

   | case | graph nodes | mem_alloc | mem_free |
   |---|---|---|---|
   | no workspace | 4 | 1 | 1 |
   | `set_workspace` | 2 | 0 | 0 |
   | `set_workspace` + `set_stream` (re-binds) | 2 | 0 | 0 |
   | `set_workspace` + raw `cublasSetStream` (no re-bind) | 4 | 1 | 1 |

   That's why the handle keeps the slice and `set_stream` and `set_workspace` share one helper.

2. **Context check.** `cublasSetWorkspace` returns success when given a pointer from another device's context, so the helper rejects that case itself (with `CUBLAS_STATUS_INVALID_VALUE`).

`device_ptr_mut` makes the handle's stream wait for any pending allocation or write of the workspace before cuBLAS can use it.

I also simplified the PR: `set_workspace` now takes `&mut self` like `set_stream`, so the `Mutex` is gone, and the helper is a free function. Both reasons are now comments in the code.

PR #600, opened 2026-09-17
Support zero and combined graph instantiation flags

joshhay95:
CUDA accepts a bitmask for graph instantiation, but the current enum parameter cannot represent zero or combined flags. Callers using default flags must drop down to `sys`.

Add `CudaStream::end_capture_with_flags(u64)` and `result::graph::instantiate_with_flags(graph, u64)`. The existing enum-based APIs remain available and delegate to these methods. Failed instantiation also releases the captured graph through its existing owner.

GPU tests cover zero, single, and combined flags, replayed output, and an incompatible-flags error followed by another capture. Both tests pass on a B300 in debug and release. Formatting, strict Clippy, packaging, and all-target checks with CUDA 11.4 (`no-std`) and 13.3 bindings pass.


chelsea0x3b:
We can also generate that enum as a bitmask using bindgen. There's already support for this in bindings generator, so I think we should go that route instead of this. thanks for looking into this

joshhay95:
> We can also generate that enum as a bitmask using bindgen. There's already support for this in bindings generator, so I think we should go that route instead of this. thanks for looking into this

Good point! I've updated the PR now, hopefully should be good now

src/driver/safe/graph.rs
```diff
@@ -60,12 +61,13 @@ impl CudaStream {
...
+        let mut graph = CudaGraph {
             cu_graph,
-            cu_graph_exec,
+            cu_graph_exec: std::ptr::null_mut(),
             stream: self.clone(),
-        }))
+        };
+        graph.cu_graph_exec = unsafe { result::graph::instantiate(cu_graph, flags) }?;
+        Ok(Some(graph))
     }
```
chelsea0x3b, line 70: Can you explain what is different here? If its the same we can just revert this change

chelsea0x3b:
looks great! Pending 1 comment and then will merge once we figure that out

src/driver/safe/graph.rs
joshhay95: It fixes a resource leak when graph instantiation fails. Previously, the code:

1. Obtained a raw CUDA graph handle.
2. Tried to create its executable version.
3. Wrapped both handles in a Rust CudaGraph, which handles cleanup automatically.

If step 2 failed, eg because the flags were incompatible, the ? returned immediately. The Rust owner was never created, so nothing destroyed the captured CUDA graph.

Might be overly cautious but seems a bit safer

joshhay95: I could refactor to this if thats more readable?

```rust
let cu_graph_exec =
    match unsafe { result::graph::instantiate(cu_graph, flags) } {
        Ok(exec) => exec,
        Err(error) => {
            self.ctx
                .record_err(unsafe { result::graph::destroy(cu_graph) });
            return Err(error);
        }
    };

Ok(Some(CudaGraph {
    cu_graph,
    cu_graph_exec,
    stream: self.clone(),
}))
```

chelsea0x3b: Ah hah! yes that refactor makes it a lot more clear, lets go with that. good catch

joshhay95: thanks! pushed up the refactor. should be ready now 🙂

